PR #1, opened 2019-02-27
Added Chun's ACO changes and fixed a typo in sched.ini

vangthao95:
In sched.ini file:
LIST_SCHED_TYPE changed to HEUR_SCHED_TYPE.

This error was found while trying to set the scheduler to ACO. It would default to LIST because it was looking for HEUR_SCHED_TYPE and not finding it. LIST is the default if none is chosen or found.


kerbowa:
Thanks for finding the bug in the example sched.ini.

PR #2, opened 2019-05-07
Added documentations on how to build flang, scripts for CPU2006, and patches for LLVM6.

vangthao95:


kerbowa:
Is the runspec wrapper an updated version, or the same that is already in the scripts folder?

vangthao95:
I have updated it to also print the weighted spills.

kerbowa:
> I have updated it to also print the weighted spills.

Can you delete the old one then? I think weighted spills are always printed when we use the wrapper.

Would you please rename the documentations directory to documentation. Thanks.

kerbowa:
We could just rename SPEC -> CPU2006 and then replace the old runspec wrapper in that directory.

kerbowa:
LGTM

PR #6, opened 2019-07-27
Remove eroneous 'O_LARGEFILE'

### LINKED ISSUE #5: Compiling for MacOS
In compiling for MacOS, I found two compile errors.
One was with `O_LARGEFILE` - which as far as I can tell is a flag for opening files, which does not exist in the MacOS libraries.
I found [this StackOverflow answer](https://stackoverflow.com/a/3425945/1291990) which seems to claim that this flag is used in the kernel, and isn't intended for use in applications.

The other issue was with the use of `constexpr`, in this declaration
```cpp
#define LSHPair std::pair<const char *, LISTSCHED_HEURISTIC>
static constexpr LSHPair HeuristicNames[] = {
    LSHPair("CP", LSH_CP),    LSHPair("LUC", LSH_LUC),
    LSHPair("UC", LSH_UC),    LSHPair("NID", LSH_NID),
    LSHPair("CPR", LSH_CPR),  LSHPair("ISO", LSH_ISO),
    LSHPair("SC", LSH_SC),    LSHPair("LS", LSH_LS),
    LSHPair("LLVM", LSH_LLVM)};
```
The error I got was
```
Constexpr variable 'HeuristicNames' must be initialized by a constant expression
```

I "fixed" these errors by simply removing the offending `O_LARGEFILE` and `constexpr`, but I have no idea what ramifications this may have.

mfdorst:
This flag does not exist on some platforms, and prevents the code from compiling on MacOS.

Closes #5.

### DIFF
--- a/README-MacOS.md
+++ b/README-MacOS.md
@@ -1,0 +1,60 @@
+# Building on MacOS with Xcode
+
+You will need Xcode, which you can find in the Mac App Store.
+
+1. Follow [these directions][quck-start] through step 4.
+
+2. Build an Xcode project
+
+```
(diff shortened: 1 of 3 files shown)


kerbowa:
LGTM

PR #7, opened 2019-07-28
Improve README

mfdorst:
Clarify which steps are for Ubuntu and which are for MacOS

### DIFF
--- a/README-MacOS.md
+++ b/README-MacOS.md
@@ -56,7 +56,7 @@ Select the `clang` scheme and then select `Product > Build` (or press `Cmd + B`)
 
 Wait for the build to complete, and repeat this with the `OptSched` scheme.
 
-## Test the build
+### Test the build
 
 Return to [step 6](README.md#Test-the-build-ubuntu-and-macos) of the quick start guide to test the build.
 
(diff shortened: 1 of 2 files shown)


kerbowa:
Is there anything special steps to build on MacOS if you don't want to use xcode?

kerbowa:
I think it's better if maybe the README has general info and we can also have install-linux/install-MacOS as separate documents. What do you think?

mfdorst:
That's a good question. It didn't occur to me to try, but I guess the instructions for building on Ubuntu should work on MacOS, provided you install `cmake` first. I was already planning on setting it up in CLion and providing alternate instructions for that - I can add building without any IDE to my todo list.

I think that's a good idea. So basically, remove the "quick start guide" from README and add two more specific readme's, perhaps `README-Ubuntu` and `README-MacOS`? And we can replace the guide with links to the different readme's. And I think I'll just copy the duplicate instructions into the MacOS readme so all the directions will be in one place.

kerbowa:
I agree that having some duplication is okay; it will be easier to read. We could have have one INSTALL.md which is linked in the main README like in most of our ROCm repos i.e. https://github.com/ROCm-Developer-Tools/HIP

mfdorst:
Okay. So move the quick-start guide to `INSTALL.md`, which will (for now) have an Ubuntu section and a MacOS section. And we will link to `INSTALL.md` in `README.md`. Sound good?

PR #12, opened 2019-07-30
Remove `constexpr` from LSHPair

### LINKED ISSUE #10: Either use C++14, or remove 'constexpr'
When I first built on MacOS I ran into a problem with [this declaration][problematic-code]. The problem was that I wasn't building with C++14 enabled, and in C++11 and earlier, `std::pair`'s constructor is not `constexpr`. I solved this by having my IDE use C++14, but the issue reared its head again today when I tried to build from the command line.
The real problem seems to be that the rest of LLVM is targeting C++11, while OptSched targets C++14 - _but we don't have that specified in the CMakeLists file_. So CMake (on some platforms, including MacOS) will default to C++11, and our code will not compile. 

We seem to have two choices:
1. Target `C++11` everywhere, and remove the offending `constexpr`, or
2. Explicitly target `C++14`.

I've made two competing branches - you can decide which one to merge.

[problematic-code]: https://github.com/CSUS-LLVM/OptSched/blob/ca7b952a572de990a3ff7218122278216718889f/lib/Wrapper/OptimizingScheduler.cpp#L48

mfdorst:
std::pair is not constexpr in C++11, only in C++14 and later.

Closes #10

### DIFF
--- a/lib/Wrapper/OptimizingScheduler.cpp
+++ b/lib/Wrapper/OptimizingScheduler.cpp
@@ -45,7 +45,7 @@ bool OPTSCHED_gPrintSpills;
 
 // An array of possible OptSched heuristic names
 #define LSHPair std::pair<const char *, LISTSCHED_HEURISTIC>
-static constexpr LSHPair HeuristicNames[] = {
+static LSHPair HeuristicNames[] = {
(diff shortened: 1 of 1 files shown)


mfdorst:
Closes #11

Issue #9, opened 2019-07-30
Fix Clang-Tidy violations?


mfdorst:
I'm personally a big fan of linter tools, and since this is an LLVM extension it seems like we should be using [Clang-Tidy](https://clang.llvm.org/extra/clang-tidy/).

I just wanted to check before I submit a pull request with a whole bunch of little code changes to fix clang-tidy issues.

Some examples of clang-tidy changes would be:
```cpp
std::map<string, string>::const_iterator it = settings.find(name);
```
> ClangTidy: Use auto when declaring iterators
```cpp
auto it = settings.find(name);
```
---
```cpp
while (!file.fail() && name.size() && name[0] == '#') {
```
> ClangTidy: The 'empty' method should be used to check for emptiness instead of 'size'
```cpp
while (!file.fail() && name.empty() && name[0] == '#') {
```

kerbowa:
Sure!

mfdorst:
Cool! What about unused code? We have a lot of unused code and I'm not sure if it's because we _will_ use it, or if it's just old dead code that hasn't been removed.

kerbowa:
There is definitely a lot of refactoring/cleanup that could be done, and I don't think anyone would object to these sort of things.

I did a clang-format passover in the past and in some of my private branches I'd removed some code that is no longer relevant if we are not doing scheduling at the SD level. It's fine to remove dead code; be careful though.

The older portions of the codebase also need to be updated in terms of the data structures used to rely on stl and LLVM more. Variables need to be renamed and follow LLVM style and CamelCase: https://llvm.org/docs/CodingStandards.html

The significant overhead from loading config files and the way we handle compiler options needs to be improved since it has a big impact on compile time.

There are lots of these tasks to be done to improve the overall code quality which is quite poor at the moment, but these types of tasks are not likely to be things that are interesting to Dr. Shobaki.

I don't have time to work on/contribute to this project at the moment, but feel free to do pretty much anything.

mfdorst:
I see a lot of variables named `likeThis_` with an underscore at the end. I noticed in one class that all the protected variables ended in underscores while the public ones did not. Is that a convention we should continue to uphold?

kerbowa:
The convention used is that class members end in an underscore. LLVM's coding style has CamelCase for both local and class members. We should also rename variables in cases where the abbreviation (wrong spelling) makes it hard to read.

mfdorst:
[...]

PR #13, opened 2019-07-31
Clang-tidy and indentation fix

mfdorst:
**Warning! This branch has a huge number of changes because I fixed the indentation on every source file in the project.** Basically anyone who's working on anything right now is going to have trouble merging after this commit.

#9


mfdorst:
For whatever reason when I fix the indentation of files it sometimes leaves comments completely unindented. I don't have the patience to search through every file for that so if people could fix those issues when they find them that'd be cool

kerbowa:
This is incorrect indentation and doesn't follow the LLVM conventions.

mfdorst:
I can't seem to get `clang-tidy` working on this project. My editor provides contextual clang-tidy warnings, but I can't figure out how to run the command on its own.
`clang-format` works, but I don't know if the default format rules are the right ones - do you have a configuration file for clang-format too?

kerbowa:
There is a clang-format config file in the same repo. I just use 'clang-format -style=llvm'. Can you just copy the dot file to the reop and run clang-tidy from the command line?

mfdorst:
clang-tidy works by basically compiling the whole source tree. So in general you have to pass it the same commands you give your compiler. CMake obviously complicates that, but you can have CMake generate a "compilation database" which tells clang-tidy how to compile your code. I tried generating a compilation database for just the OptSched target, but it doesn't seem to like that. For now I'm just gonna switch to clang-format until I can figure out how to work with clang-tidy.

kerbowa:
https://github.com/CSUS-LLVM/OptSched/blob/c980bbf445e14111404ab20ccb018a68dd03e042/CMakeLists.txt#L24

I added compilation db by default already. I'm pretty sure llvm does the same. I think maybe it's your ide that's wonky, can't really help you there.

mfdorst:
Hold up don't merge this just yet, I managed to delete `aco.cpp`

kerbowa:
Was going to ask :)

I think I will take a closer look tomorrow. Why is it marking some class methods as static?

mfdorst:
Is clang-format doing that? Or are you asking about the earlier clang-tidy commits?

mfdorst:
clang-tidy does sometimes recommend a method be static - I think I only made that change once though - I'll take a closer look in case that was a bad idea.

kerbowa:
I think some things from clang-tidy are still here like, adding override for example.

mfdorst:
The last two commits have no clang-tidy fixes. The previous ones do, with the exception of 'Fix indentation project-wide'.

mfdorst:
Do you want me to make a pull request with _only_ clang-format changes, and a different one for clang-tidy changes?

mfdorst:
Actually I like that idea. I will do that tomorrow - I have math homework to do right now.

mfdorst:
Closing this. #18 runs clang-format for the whole project. If that gets merged, I'll rebase these changes on top of those which should make everything cleaner and easier to follow.

PR #14, opened 2019-07-31
Updated hot funcs ini for LLVM6

vangthao95:
Some functions were renamed in LLVM6. Updated to work with both LLVM3.9 and LLVM6.


kerbowa:
Don't want to nit pick too much, but do you think we should have two hotfuncs files i.e. one for each version, because the unused lines are not doing anything right. I wonder if we could just demangle the function names so this isn't an issue.

vangthao95:
I somehow manage to forget that we already have a specific repo for LLVM3.9 so we don't really need the functions for LLVM3.9 in this repo. I'll just remove those.

PR #17, opened 2019-07-31
Add .clang-format and .clang-tidy

kerbowa:
Add code style/linter configuration files.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,2 @@
+BasedOnStyle: LLVM
+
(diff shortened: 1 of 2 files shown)


kerbowa:
Would add mfdorst as reviewer but needs to be part of CSUS_LLVM.

mfdorst:
Your `.clang-format` file is empty - I believe you want to run
```
clang-format -style=llvm -dump-config > .clang-format
```

mfdorst:
But then again, since we're just using LLVM style, we can just run
```
clang-format -i -style=llvm **/*.cpp **/*.h
```
so the file may not be necessary. I was thinking of just making a tiny script for that to make it easier for others, something like `clang-tidy.sh` and we can have everyone run that before committing.

kerbowa:
I see `BasedOnStyle: LLVM`, which is equivalent to `-style=llvm`.

mfdorst:
Is it? You may be right. I assumed that was just a comment. LGTM.

PR #19, opened 2019-08-01
Fix clang-tidy warnings

mfdorst:
Add fixme annotations for clang-tidy violations which need attention

Same changes as #13, but rebased on top of 65ea3b2fcd52853198f4016fc206a58913640439.


mfdorst:
I undid the change that made a method static - I think clang-tidy suggests it any time a method doesn't use any class members, but it's hard to tell between methods that _should_ be static and methods that just aren't fully implemented yet.

kerbowa:
I think that this needs to be tested first.

kerbowa:
Closing for now since this appears to be dead. Lets address these types of changes in smaller PRs like #39.
My worry is that we introduce bugs that are hard to triage with so many unrelated changes.

PR #21, opened 2019-10-31
Fix bug in adjusted use calculation.

vangthao95:
For loop would terminate early if useCnt was deducted. This was causing incorrect adjusted use counts for some instructions.

### DIFF
--- a/lib/Scheduler/sched_basic_data.cpp
+++ b/lib/Scheduler/sched_basic_data.cpp
@@ -698,12 +698,13 @@ bool SchedInstruction::ProbeScsrsCrntLwrBounds(InstCount cycle) {
 void SchedInstruction::ComputeAdjustedUseCnt_() {
   Register **uses;
   int useCnt = GetUses(uses);
+  int tempAdjustedUseCnt = useCnt;
 
   for (int i = 0; i < useCnt; i++) {
     if (uses[i]->IsLiveOut())
-      useCnt--;
+      tempAdjustedUseCnt--;
   }
-  adjustedUseCnt_ = useCnt;
+  adjustedUseCnt_ = tempAdjustedUseCnt;
 }
 
 InstCount SchedInstruction::GetFileSchedOrder() const {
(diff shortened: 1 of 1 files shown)


lib/Scheduler/sched_basic_data.cpp
```diff
@@ -698,12 +698,13 @@ bool SchedInstruction::ProbeScsrsCrntLwrBounds(InstCount cycle) {
 void SchedInstruction::ComputeAdjustedUseCnt_() {
   Register **uses;
   int useCnt = GetUses(uses);
+  int tempAdjustedUseCnt = useCnt;
```
kerbowa: NITPICK: You could just write:
```suggestion
  adjustedUseCnt_ = GetUses(uses);;
```

lib/Scheduler/sched_basic_data.cpp
```diff
@@ -698,12 +698,13 @@ bool SchedInstruction::ProbeScsrsCrntLwrBounds(InstCount cycle) {
...
 
   for (int i = 0; i < useCnt; i++) {
     if (uses[i]->IsLiveOut())
-      useCnt--;
+      tempAdjustedUseCnt--;
```
kerbowa:
```suggestion
     adjustedUseCnt_--;
```

lib/Scheduler/sched_basic_data.cpp
```diff
@@ -698,13 +698,12 @@ bool SchedInstruction::ProbeScsrsCrntLwrBounds(InstCount cycle) {
 void SchedInstruction::ComputeAdjustedUseCnt_() {
   Register **uses;
   int useCnt = GetUses(uses);
-  int tempAdjustedUseCnt = useCnt;
+  adjustedUseCnt_ = GetUses(uses);
```
kerbowa: Sorry, lets also avoid calling the function twice.
`adjustedUseCnt_ = useCnt;`

PR #22, opened 2019-10-31
Update conditions to determine if an instruction feeds a user.

vangthao95:
Current instruction feeds a user if:
1.) It has a successor that has an adjusted use count greater than 0 (Use register is not live-out and will eventually be closed).
2.) The successor instruction does not define more registers than it uses.

### DIFF
--- a/lib/Scheduler/data_dep.cpp
+++ b/lib/Scheduler/data_dep.cpp
@@ -3260,16 +3260,16 @@ bool DataDepGraph::DoesFeedUser(SchedInstruction *inst) {
     SchedInstruction *succInst = static_cast<SchedInstruction *>(succ);
     
     int curInstAdjUseCnt = succInst->GetAdjustedUseCnt();
-    // Ignore successor instructions that does not close registers
+    // Ignore successor instructions that does not close live intervals
     if (curInstAdjUseCnt == 0)
       continue;
-    // Ignore instructions that define more registers than it closes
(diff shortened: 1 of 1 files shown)


lib/Scheduler/data_dep.cpp
```diff
@@ -3258,19 +3258,20 @@ bool DataDepGraph::DoesFeedUser(SchedInstruction *inst) {
   for (GraphNode *succ = rcrsvSuccs->GetFrstElmnt(); succ != NULL;
        succ = rcrsvSuccs->GetNxtElmnt()) {
     SchedInstruction *succInst = static_cast<SchedInstruction *>(succ);
-    Register **uses;
-    int numUses = succInst->GetUses(uses);
+    
+    int curInstAdjUseCnt = succInst->GetAdjustedUseCnt();
+    // Ignore successor instructions that does not close registers
```
kerbowa: NITPICK: Normally we say "opens/closes live intervals".

lib/Scheduler/data_dep.cpp
```diff
@@ -3258,19 +3258,20 @@ bool DataDepGraph::DoesFeedUser(SchedInstruction *inst) {
...
+    if (curInstAdjUseCnt == 0)
+      continue;
+    // Ignore instructions that define more registers than it closes
```
kerbowa: Ditto

PR #24, opened 2019-11-21
Fix schedule specific region setting.

vangthao95:
Schedule specific region setting was skipping the regions listed instead of scheduling them only. This fixes the bug and only schedules the region specified if the setting is enabled.


vangthao95:
The find function returns "std::end(regionList)" if a region isn't in the list to be scheduled so this change does make it return true for all regions except the ones to schedule. The original was returning true for all regions listed to be scheduled and was causing them to be skipped.

kerbowa:
I'm just confused at what point this got borken because this was definitely working correctly in the past.

PR #26, opened 2019-11-28
[Two-Pass] Integrated two-pass implementation from AMDGPU to base OptSched scheduler.

vangthao95:
This change allows the base OptSched scheduler to also use the two-pass approach that is implemented in the AMDGPU target scheduler.

There are a few changes that have been added:
- Added more settings that can be set for the two-pass scheduler.
    1.) Different timeouts for each pass
    2.) Different enumerator heuristic for each pass
- Moved region recording to base OptSched scheduler class to avoid code redundancy.
- Moved occupancy pass and ILP pass source code to base OptSched scheduler class since both use the same code (most of it was just changing the settings).

Changes were tested with plaidbench running both the original and the new changes 3 times each. Results are identical when taking into account random variation. 
- Compile-time: +0.03% difference
- TIles per second: -0.12% difference
- Examples per second: +0.13% difference

Note: With these changes, to use the AMDGPU scheduler, USE_TWO_PASS must be enabled and the settings should be set with the new settings.


lib/Wrapper/OptimizingScheduler.cpp
```diff
@@ -564,6 +592,21 @@ bool ScheduleDAGOptSched::isOptSchedEnabled() const {
...
 
+bool ScheduleDAGOptSched::isTwoPassEnabled() const {
+  // check scheduler ini file to see if two pass scheduling is enabled
+  auto twoPassOption =
+      SchedulerOptions::getInstance().GetString("USE_TWO_PASS");
+  if (twoPassOption == "YES") {
+    return true;
+  } else if (twoPassOption == "NO") {
+    return false;
+  } else {
```
kerbowa: llvm_unreachable

lib/Wrapper/OptimizingScheduler.cpp
```diff
@@ -564,6 +592,21 @@ bool ScheduleDAGOptSched::isOptSchedEnabled() const {
   }
 }
 
+bool ScheduleDAGOptSched::isTwoPassEnabled() const {
+  // check scheduler ini file to see if two pass scheduling is enabled
+  auto twoPassOption =
+      SchedulerOptions::getInstance().GetString("USE_TWO_PASS");
+  if (twoPassOption == "YES") {
```
kerbowa: NITPICK: LLVM style has no brackets with one line conditions

lib/Wrapper/OptimizingScheduler.cpp
```diff
@@ -707,6 +786,52 @@ void ScheduleDAGOptSched::finalizeSchedule() {
...
+  LLVM_DEBUG(dbgs() << "Starting second pass through\n");
+  SecondPass = true;
+  LatencyPrecision = LTP_ROUGH;
+  // Set times for the second pass
+  RegionTimeout = SecondPassRegionTimeout;
+  LengthTimeout = SecondPassLengthTimeout;
+  // Set the heuristic for the enumerator in the second pass.
+  EnumPriorities = SecondPassEnumPriorities;
+  // Force the input to the balanced scheduler to be the sequential order of the
+  // (hopefully) good max occupancy schedule. We don’t want the list scheduler
```
kerbowa: Occupancy doesn't make sense in a generic context.

lib/Wrapper/OptimizingScheduler.cpp
vangthao95: Does it return an error or are you suggesting to call llvm_unreachable("<Error log>") instead of an else statement?

kerbowa: I think we should have errors for unrecognized options. The message right now will only print with the debug flag passed to llvm. I think it's to easy to make mistakes with sched.ini options so we should tighten things up.

https://github.com/llvm/llvm-project/blob/905b002c139f039a32ab9bf1fad63d745d12423f/llvm/lib/Target/AMDGPU/SIInstrInfo.cpp#L4117

vangthao95: I agree. I have changed it to output an error in the latest commit.

lib/Wrapper/OptimizingScheduler.cpp
vangthao95: Fixed

lib/Wrapper/OptimizingScheduler.cpp
vangthao95: Fixed

kerbowa:
LGTM

PR #28, opened 2019-12-18
[AMDGPU] New plaidbench scripts for stats collection

vangthao95:


vangthao95:
> LGTM
> It would be good for someone to create a workload agnostic version of these sorts of tools eventually. On the CPU we are pretty focused on SPEC, but I imagine there will be a lot more variety of GPU benchmarks.

I agree. I will look to see if this can be done easily since our scheduler should always output the same information regardless of targets or benchmarks.

PR #30, opened 2019-12-22
Updated run-plaidbench.sh to have a configurable number of runs.

addictgamer:
What it says on the tin.

The only issue I noticed was that the extract script uses `plaidbench-optsched-1` as the first run, and puts `plaidbench-optsched-0` at the end of columns.

### DIFF
--- a/util/plaidbench/run-plaidbench.sh
+++ b/util/plaidbench/run-plaidbench.sh
@@ -19,13 +19,30 @@ BatchSize=16
 Command="plaidbench --examples $Examples --batch-size $BatchSize --results "
 # Note: The run number at the end such as "01" should be kept. This number is used
 # in other scripts. 
-DirectoryName="plaidbench-optsched-01"
+DirectoryBasePath="plaidbench-optsched-"
 Keras="keras --no-fp16 --no-train"
 
-for i in 1
+NumArgs="$#"
+NumIterations=1
+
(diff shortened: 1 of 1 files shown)


vangthao95:
Has this been tested for compatibility with other scripts? I believe extract-plaidbench-data.py uses the last two characters to label the run number (IE run 01, 02, 03 and so on) when exporting the data to a spreadsheet, which is sloppy and can easily be changed, but I would like to maintain current compatibility.

addictgamer:
I only tested it with `extract-plaidbench-data.py`, which works aside from the aforementioned issue. Regarding this issue, I can take a look at updating the `extract-plaidbench-data.py` script -- an immediate and simple solution would be to delimit the filename on '-' and use the last chunk for labeling the run number.

I can give the other scripts a run when I have some time.

addictgamer:
Ah, I just noticed that it does copy over the - in the column labels, so it shows "Run -1", "Run -2", "Run -0". So even if the `extract-plaidbench-data.py` modification I mentioned doesn't fix the first issue, it would at least fix this one.

vangthao95:
The other scripts should be fine since they do not use a run number. I'll make a modification to the script to fix the "Run -1" issue. Besides that this looks good to me.

addictgamer:
Merged.

PR #31, opened 2019-12-31
[Scripts] Added schedule length and occupancy scripts. Improved older scripts.

vangthao95:
All scripts can now output a spreadsheet (enabled by default). This function can be disabled with the "--disable" option. Added "-v" option to display the stats processed to the terminal. 

Additionally, scripts do not have to be in the same directory anymore. Use the "-i" option to pass in the path to a directory to process.

Fixed inconsistent use of single/double quotes. Fixed and added some more comments.

Note: There have been no changes to the following file besides eliminating quotes inconsistency. 
> plaidbench-validation-test.py


util/plaidbench/get-optsched-stats.py
```diff
@@ -1,242 +1,358 @@
...
+
+def writeBenchmarkNames(ws, row):
+    for bench in benchmarks:
+        ws['A' + str(row)] = bench
+        row += 1
+    ws['A' + str(row)] = 'Overall'
+    ws['A' + str(row)].font = Font(bold=True)
+
+def createSpreadsheets(output):
+    if 'xls' not in output[-4:]:
```
addictgamer: should this be xls or xlsx?

vangthao95: It should be xls but it can also detect xlsx. Both are valid formats for excel spreadsheets. xls is an older format and I only kept xls because the xlwt module used in 'extract-plaidbench-data.py' saves spreadsheets in xls format. I just recently found out that xlwt is no longer being updated and can swap to openpyxl, which uses the latest format, if this is an issue.

addictgamer: Ah, ok. Got it. Was just wondering about it because the line below says xlsx.

util/plaidbench/extract-plaidbench-data.py
```diff
@@ -22,135 +22,157 @@
...
+    ./extract-plaidbench-data.py -i /home/tom/plaidbench-runs
+    
+    where plaidbench-runs/ contains
+        plaidbench-optsched-01/
+        plaidbench-optsched-02/
+        ...
+        plaidbench-amd-01/
+        ...
 '''
 #!/usr/bin/python3
```
addictgamer: This should be at the top of the file.

addictgamer: Same for the rest of the files.

vangthao95: Thanks for the catch. Changes made.

PR #33, opened 2020-01-25
Fix a memory leak issue.

vangthao95:


vangthao95:
> Thanks good find! Can we make region a unique_ptr?

Updated to make region a unique_ptr. I have tested that this also solves the memory leak issue.

lib/Wrapper/OptimizingScheduler.cpp
```diff
@@ -379,10 +379,12 @@ void ScheduleDAGOptSched::schedule() {
   addGraphTransformations(BDDG);
 
   // create region
-  SchedRegion *region = new BBWithSpill(
-      OST.get(), static_cast<DataDepGraph *>(DDG.get()), 0, HistTableHashBits,
-      LowerBoundAlgorithm, HeuristicPriorities, EnumPriorities, VerifySchedule,
-      PruningStrategy, SchedForRPOnly, EnumStalls, SCW, SCF, HeurSchedType);
+  std::unique_ptr<SchedRegion> region =
```
kerbowa: Should use llvm::make_unique

vangthao95:
Would it be like this? I see that our code uses this format for a unique_ptr.
```c++
  std::unique_ptr<SchedRegion> region = llvm::make_unique<BBWithSpill>(
      OST.get(), static_cast<DataDepGraph *>(DDG.get()), 0, HistTableHashBits,
      LowerBoundAlgorithm, HeuristicPriorities, EnumPriorities, VerifySchedule,
      PruningStrategy, SchedForRPOnly, EnumStalls, SCW, SCF, HeurSchedType);
```

Also, how about using `auto` for region?

kerbowa:
> Would it be like this? I see that our code uses this format for a unique_ptr.
> 
> ```c++
>   std::unique_ptr<SchedRegion> region = llvm::make_unique<BBWithSpill>(
>       OST.get(), static_cast<DataDepGraph *>(DDG.get()), 0, HistTableHashBits,
>       LowerBoundAlgorithm, HeuristicPriorities, EnumPriorities, VerifySchedule,
>       PruningStrategy, SchedForRPOnly, EnumStalls, SCW, SCF, HeurSchedType);
> ```
> 
> Also, how about using `auto` for region?

That looks correct. Using auto is preferable thanks.

Note that in later builds LLVM migrated to using std::make_unique from c++ 14. When porting the code to trunk this is one of the things I needed to change. Since we are working with an older version of ROCm llvm::make_unique is correct.

vangthao95:
> > Would it be like this? I see that our code uses this format for a unique_ptr.
> > ```c++
> >   std::unique_ptr<SchedRegion> region = llvm::make_unique<BBWithSpill>(
> >       OST.get(), static_cast<DataDepGraph *>(DDG.get()), 0, HistTableHashBits,
> >       LowerBoundAlgorithm, HeuristicPriorities, EnumPriorities, VerifySchedule,
> >       PruningStrategy, SchedForRPOnly, EnumStalls, SCW, SCF, HeurSchedType);
> > ```
> > 
> > 
> > Also, how about using `auto` for region?
> 
> That looks correct. Using auto is preferable thanks.
> 
> Note that in later builds LLVM migrated to using std::make_unique from c++ 14. When porting the code to trunk this is one of the things I needed to change. Since we are working with an older version of ROCm llvm::make_unique is correct.

Thank you for the clarification. I have updated region to use `auto` and `llvm::make_unqiue`.

PR #40, opened 2020-02-20
[Two-Pass] Fix schedule specific regions bug

vangthao95:
Enabling the setting "SCHEDULE_SPECIFIC_REGIONS" causes regions that are not selected for scheduling to bypass the recording phase. This messes up the region numbering in the multiple passes approach and causes a bug where we would schedule incorrect or no regions at all. This patch makes it so that we record all regions even if they are going to be skipped when scheduling to keep proper numbering of the regions.

### DIFF
--- a/lib/Wrapper/OptimizingScheduler.cpp
+++ b/lib/Wrapper/OptimizingScheduler.cpp
@@ -130,7 +130,7 @@ static bool scheduleSpecificRegion(const StringRef RegionName, const Config &Sch
       SchedIni.GetBool("SCHEDULE_SPECIFIC_REGIONS");
 
   if (!ScheduleSpecificRegions)
-    return false;
+    return true;
 
   const std::list<std::string> RegionList =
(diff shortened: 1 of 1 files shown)


lib/Wrapper/OptimizingScheduler.cpp
```diff
@@ -125,17 +125,17 @@ nextIfDebug(MachineBasicBlock::iterator I,
   return I;
 }
 
-static bool skipRegion(const StringRef RegionName, const Config &SchedIni) {
+static bool scheduleSpecificRegion(const StringRef RegionName, const Config &SchedIni) {
   const bool ScheduleSpecificRegions =
       SchedIni.GetBool("SCHEDULE_SPECIFIC_REGIONS");
 
   if (!ScheduleSpecificRegions)
     return false;
```
kerbowa: true

vangthao95: Thanks for catching this!

PR #41, opened 2020-02-22
Disable RP-only graph transformations in balanced mode

kerbowa:


Quincunx271:
Alternatively, would it be better to simply not add the graph transformations over at `addGraphTransformations()` when `LatencyPrecision != LTP_UNITY`? That's what I implemented on my branch.

kerbowa:
> Alternatively, would it be better to simply not add the graph transformations over at `addGraphTransformations()` when `LatencyPrecision != LTP_UNITY`? That's what I implemented on my branch. That way, it's not possible to forget to disable the graph transformations when setting up the other sched.ini settings for a pass.

That would work, although not all graph transformations are necessarily only valid in RP-only mode. For example LLVM's DAG mutations don't have this restriction. The only thing I would request is that there is some warning that static node sup is being disabled if it is detected that LLVM latencies are enabled at the same time. Because I think this warning would be good, both changes should go in to reflect the fact that with multi-pass we are intentionally disabling graph transformations and it is not some user error. We could even just assert in addGraphTransformations if StaticNodeSup and non-UNITY latencies are enabled.

Quincunx271:
Indeed, it's specific to the particular graph transformations algorithm. The code I added includes a log statement when it skips the particular transformation. Other graph transformations added in `addGraphTransformations()` could choose whether to be added independently of the static node superiority algorithm.

Issue #47, opened 2020-02-28
General Refactoring Planning


kerbowa:
Recent discussion in #45 and elsewhere has made it clear that we should make refactoring a priority. We have limited resources, however I believe that a strong argument can be made that these changes will help the project grow and have a positive net effect.

It seems like most active contributors feel similarly, so lets prioritize some of the idea's, and then decide how much time should be invested into this.

### Why Refactor

This is an open source project. We would like to increase the number of collaborators. Companies have contacted us wanting to try our these scheduling techniques only to be discouraged by the lack of usability. The open source community respects well maintained code, and if the goal is to invite more collaboration we need to have things like easy-of-use, documentation, CI, ect in place.

The lack of any real testing infrastructure is a big problem. If the end goal is to have this be a tool recognized by the LLVM community, like say Poly for example, then this type of refactoring work is how we should start along that path.

### What work needs to be done

1. Testing infrastructure. We should decide on lit, filecheck, gtest or some combination. In my opinion this is the most critical missing piece right now. It's impossible for new people to navigate the project without unit tests which can show them how the pieces of the scheduler work together.

2. Refactor scheduling pass manager. Each scheduling pass should be independent. There should be classes which represent a scheduling pass and the code should not rely on flags to change behavior for different passes.

3. The older code needs to be modernized. That means smart pointers, and variable names that are descriptive. It should be written with C++14 and LLVM programming style in mind. We should use iterators/vectors/maps ect and leverage the LLVM infrastructure whenever possible.

4. Use tablegen to generate the sched.ini options. The way we handle scheduler settings is horrible. It should be possible for all flags to be added on the CL as well as read from sched.ini. We also need to have sane defaults so that things work without messing with ini files.

I hope we can create some github issues with tasks that will make contributing to this project much easier for new people.

Quincunx271:
I've also noticed that we have a lot of protected member variables. It would be much easier to follow the code if these were private variables; that way, I can know what can and can't be changed at a given point of time.

Quincunx271:
I would also like to suggest renaming of functions and variables. `UpdtRdyLst_` is not much shorter than `UpdateRdyList_` or `UpdateReadyList_`, but those are much easier to read. There's a pervasive removal of vowels: `Dynmc`, `PrirtyLst`, `RcrsvScsr`.

[...]

PR #49, opened 2020-03-11
Replace raw pointers with unique_ptr and vector

Quincunx271:
A very-much-WIP branch for replacing raw pointer arrays with either `unique_ptr<T[]>` or `vector<T>`.

I can say that this currently compiles, but I haven't run anything with it. See #47 . There's still a lot of work to do here and I'm not prioritizing this, but I figured it would be good to PR this to allow collaboration / prevent duplicate work. Feel free to push more related changes to this branch.

----

I'm preferring using `vector<T>` when possible, but if we are unsure if it is possible, `unique_ptr<T[]>` allows us to more closely replace the raw pointer.

### DIFF
--- a/include/opt-sched/Scheduler/aco.h
+++ b/include/opt-sched/Scheduler/aco.h
@@ -9,6 +9,8 @@ Last Update:  Jan. 2020
 #ifndef OPTSCHED_ACO_H
 #define OPTSCHED_ACO_H
 
+#include <vector>
+
 #include "opt-sched/Scheduler/gen_sched.h"
 
 namespace llvm {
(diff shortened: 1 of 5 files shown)


paul-mchugh:
I have a commit coming that will change the raw pointers in aco.cpp and fix other bugs.  The aco changes you are making are redundant.

kerbowa:
In most cases std::vector should be replaced with [llvm::small_vector](http://llvm.org/docs/ProgrammersManual.html#llvm-adt-smallvector-h).

This is standard in LLVM patches.

Quincunx271:
Closing this. I'll open a new PR when I'm more actively working on this

PR #50, opened 2020-03-13
Compare absolute costs

Quincunx271:
Many of our scripts were comparing relative costs rather than absolute costs. This interferes with analyzing graph transformations, because the graph transformations are run before the lower bound calculation. This means that the absolute reference point varies depending on whether graph transformations have been run.

To fix this, I'm unconditionally logging the lower bound and changing the scripts to compare absolute costs rather than relative costs.

-----

Did I miss any scripts?


vangthao95:
LGTM

PR #51, opened 2020-03-14
Add unittest infrastructure

Quincunx271:
See #47 .

This attempts to copy the setup from llvm/clang tests and unit tests. We
This setup should enable us to add regression tests later should we so
desire. See llvm/docs/TestingGuide.rst.

To run the tests, run `ninja check-optsched-unit` or whatever build
system was generated by CMake. See the BUILD.md file for more
information on running the tests.

As a minimal test-case, add tests for
`Utilities::clcltBitsNeededToHoldNum()`.


vangthao95:
LGTM

PR #52, opened 2020-03-15
Fix test setup to link with OptSched

Quincunx271:
Add tests for Config. Also, link OptSched to the test binaries.


Quincunx271:
If nobody has any objections, I would like to merge this. The prior PR which merged unit tests left out linking, since it only tested an inline function.

PR #53, opened 2020-03-16
Aco cleanup

paul-mchugh:
Changes ACO to use llvm::SmallVector/std::unique_ptr instead of the raw pointers, fixes small memory leak, adds more imformation to ACO's log output.


include/opt-sched/Scheduler/aco.h
```diff
@@ -29,22 +31,23 @@ class ACOScheduler : public ConstrainedScheduler {
...
   pheremone_t &Pheremone(InstCount from, InstCount to);
   double Score(SchedInstruction *from, Choice choice);
 
   void PrintPheremone();
 
-  SchedInstruction *SelectInstruction(std::vector<Choice> ready,
-                                      SchedInstruction *lastInst);
+  SchedInstruction *
+  SelectInstruction(const llvm::SmallVectorImpl<Choice> &ready,
+                    SchedInstruction *lastInst);
```
Quincunx271: I'm not familiar with LLVM best practices. Is `llvm::SmallVectorImpl` a better choice than `llvm::ArrayRef`?

lib/Scheduler/aco.cpp
```diff
@@ -169,21 +159,23 @@ SchedInstruction *ACOScheduler::SelectInstruction(std::vector<Choice> ready,
   return ready.back().inst;
 }
 
-InstSchedule *ACOScheduler::FindOneSchedule() {
+std::unique_ptr<InstSchedule> ACOScheduler::FindOneSchedule() {
   SchedInstruction *lastInst = NULL;
-  InstSchedule *schedule = new InstSchedule(machMdl_, dataDepGraph_, true);
+  std::unique_ptr<InstSchedule> schedule(
+      new InstSchedule(machMdl_, dataDepGraph_, true));
```
Quincunx271: Any reason not to use `llvm::make_unique<InstSchedule>(...)`?

lib/Scheduler/aco.cpp
```diff
@@ -269,26 +263,23 @@ FUNC_RESULT ACOScheduler::FindSchedule(InstSchedule *schedule_out,
...
-    InstSchedule *iterationBest = NULL;
+    std::unique_ptr<InstSchedule> iterationBest;
     for (int i = 0; i < ants_per_iteration; i++) {
-      InstSchedule *schedule = FindOneSchedule();
+      std::unique_ptr<InstSchedule> schedule = FindOneSchedule();
       if (print_aco_trace)
-        PrintSchedule(schedule);
-      if (iterationBest == NULL ||
+        PrintSchedule(schedule.get());
+      if (iterationBest == nullptr ||
```
Quincunx271: How do we feel about just `!iterationBest`? That's what the CppCoreGuidelines says to do: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es87-dont-add-redundant--or--to-conditions

lib/Scheduler/aco.cpp
```diff
@@ -297,29 +288,32 @@ FUNC_RESULT ACOScheduler::FindSchedule(InstSchedule *schedule_out,
     /* PrintSchedule(iterationBest); */
     /* std::cout << iterationBest->GetCost() << std::endl; */
     // TODO DRY
-    if (bestSchedule == NULL ||
+    if (bestSchedule == nullptr ||
```
Quincunx271: See above

lib/Scheduler/aco.cpp
```diff
@@ -439,7 +433,8 @@ void PrintSchedule(InstSchedule *schedule) {
 
 void ACOScheduler::setInitialSched(InstSchedule *Sched) {
   if (Sched) {
-    InitialSchedule = new InstSchedule(machMdl_, dataDepGraph_, VrfySched_);
+    InitialSchedule = std::unique_ptr<InstSchedule>(
+        new InstSchedule(machMdl_, dataDepGraph_, VrfySched_));
```
Quincunx271: See above

lib/Scheduler/aco.cpp
```diff
@@ -257,8 +250,9 @@ FUNC_RESULT ACOScheduler::FindSchedule(InstSchedule *schedule_out,
   for (int i = 0; i < pheremone_size; i++)
     pheremone_[i] = 1;
   initialValue_ = 1;
+  std::unique_ptr<InstSchedule> heuristicSched = FindOneSchedule();
```
vangthao95: Suggest the use of `auto`. https://llvm.org/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable

lib/Scheduler/aco.cpp
```diff
@@ -269,26 +263,23 @@ FUNC_RESULT ACOScheduler::FindSchedule(InstSchedule *schedule_out,
...
   Config &schedIni = SchedulerOptions::getInstance();
   int noImprovementMax = schedIni.GetInt("ACO_STOP_ITERATIONS");
   int noImprovement = 0; // how many iterations with no improvement
   int iterations = 0;
   while (true) {
-    InstSchedule *iterationBest = NULL;
+    std::unique_ptr<InstSchedule> iterationBest;
     for (int i = 0; i < ants_per_iteration; i++) {
-      InstSchedule *schedule = FindOneSchedule();
+      std::unique_ptr<InstSchedule> schedule = FindOneSchedule();
```
vangthao95: Same as above.

lib/Scheduler/aco.cpp
paul-mchugh: The guidelines say that auto should be used when it improves readability. One of the circumstances where it improves readability that the standards identified was when the type of the object is obvious from context. (e.g. when you are assigning the result of a cast operation to a variable).  He it is not obvious that FindOneSchedule() returns a unique_ptr to an InstSchedule.  I think that use of auto here would make the code less clear and less readable.

lib/Scheduler/aco.cpp
paul-mchugh: The LLVM Coding standards don't say anything either way, and either way you have to add another operator `!` vs `== nullptr`.  ES87 only seems to cover causes where you are checking if it is != null, not causes where you are checking if it is null.

lib/Scheduler/aco.cpp
paul-mchugh: The project is currently on llvm 6, and llvm 6 uses c++11.  There is no std::make_unique c++11 (it is introduced in c++14), so we would have to use the identical llvm::make_unique.  In newer versions of llvm that use c++14 they remove llvm::make_unique.  I used the unique_ptr constructor, so we will not have to change this when we eventually update to the newer version of llvm.

include/opt-sched/Scheduler/aco.h
paul-mchugh: http://llvm.org/docs/ProgrammersManual.html says to use "Prefer to use SmallVectorImpl<T> as a parameter type."  It lets you pass SmallVector s of any size.

lib/Scheduler/aco.cpp
vangthao95: That's understandable. I'm fine with not using auto here then.

include/opt-sched/Scheduler/aco.h
Quincunx271: So does ArrayRef. https://lists.llvm.org/pipermail/llvm-dev/2019-May/132127.html

lib/Scheduler/aco.cpp
Quincunx271: It's trivial to find and replace `llvm::make_unique` with `std::make_unique` when we're ready to upgrade.

kerbowa: Right, the latest LLVM uses std::make_unique and c++14. llvm::make_unique was used with LLVM 6.0.

My understanding was that we use LLVM 6 because of flang. There are multiple projects underway to replace flang e.g. https://github.com/flang-compiler/f18

include/opt-sched/Scheduler/aco.h
```diff
@@ -29,22 +31,23 @@ class ACOScheduler : public ConstrainedScheduler {
...
   void UpdatePheremone(InstSchedule *schedule);
-  InstSchedule *FindOneSchedule();
```
kerbowa: Why is FindOneSchedule() called during the definition?

include/opt-sched/Scheduler/aco.h
```diff
@@ -29,22 +31,23 @@ class ACOScheduler : public ConstrainedScheduler {
...
-  pheremone_t *pheremone_;
+  std::unique_ptr<InstSchedule> FindOneSchedule();
+  llvm::SmallVector<pheremone_t, 0> pheremone_;
```
kerbowa: Is there any reasonable guess about the average size of pheremone or does it really make sense to have no space reserved?

lib/Scheduler/aco.cpp
kerbowa: Should be llvm::make_unique I guess.

lib/Scheduler/aco.cpp
```diff
@@ -115,7 +105,7 @@ SchedInstruction *ACOScheduler::SelectInstruction(std::vector<Choice> ready,
 
   if (RandDouble(0, 1) < choose_best_chance) {
     if (print_aco_trace)
-      std::cerr << "choose_best, use fixed bais: " << use_fixed_bias << "\n";
+      std::cerr << "choose_best, use fixed bias: " << use_fixed_bias << "\n";
     pheremone_t max = -1;
     Choice maxChoice;
     for (auto choice : ready) {
```
kerbowa: `auto &choice` to avoid copy of the struct.

paul-mchugh: missed that, will change

include/opt-sched/Scheduler/aco.h
paul-mchugh: pheremone_ has a size of (instruction count)*(instruction count+1), It is usually a largeish allocation.  It is highly variable in size depending on the scheduling region.

lib/Scheduler/aco.cpp
paul-mchugh: I will change it over to use llvm::make_unique

include/opt-sched/Scheduler/aco.h
paul-mchugh: I don't understand the question, in this LOC FindOneSchedule() is being declared, not called.

include/opt-sched/Scheduler/aco.h
paul-mchugh: I don't think SelectInstruction() s parameter types really matter, It is a private method which is called only once, and is seems unlikely that we would ever want to pass another array type instead of llvm::SmallVector.  I will change it because it seems like you are right as to the standards question.

paul-mchugh:
Unless there is an objection, I will merge this pr in.

include/opt-sched/Scheduler/aco.h
Quincunx271: If we have an idea of the distribution in scheduling region size, it could make sense to provide the small size as @kerbowa says. For example, if a block size of 4 instructions was relatively common, providing a small size of `4 * 5 == 20` would remove the allocation for regions of size 4 or smaller.

In this case, I don't think we need to worry about this. The allocation is unlikely to be a bottleneck in our code, and we can't know if it is unless we have profiled our code for small scheduling regions. Since we have not done so, I think we should leave it as `0`; potential future optimization work could use a small size of `0` as a sign that there may be some optimizing opportunity.

vangthao95:
There is no objection from me if this build has been tested for regressions.

paul-mchugh:
forgot to respond to that comment.  I have tested the code

include/opt-sched/Scheduler/aco.h
kerbowa: I agree it's unlikely to be a bottleneck, so lets leave it as is.
For awareness, in LLVM we normally pre-allocate some amount of data in the SmallVector object even if you are just guessing. The convention is to use powers of 2. `4` `8` `16` `32` ect..

PR #54, opened 2020-03-16
Add Superbuild

### LINKED ISSUE #48: Easier cloning / building
Currently, [BUILD.md](https://github.com/CSUS-LLVM/OptSched/blob/master/BUILD.md) and [documentation/flang/README.md](https://github.com/CSUS-LLVM/OptSched/blob/master/documentation/flang/README.md) list a large number of steps which require manual interaction interspersed with waiting for compilation to finish. Ideally, we could just do a single download step and a single build step (possibly with a single prior configure step).

I made a master repository which allows for this via git submodules: https://github.com/Quincunx271/OptSchedProject . A recursive git clone grabs all the source code needed plus OptSched itself, and there's a super-build CMakeLists.txt to kick off the build for everything. You can also clone only LLVM + OptSched if flang is not desired.

The idea is to have CSUS-LLVM/OptSchedProject and associated git repositories to make setting up OptSched easier. The associated repositories are the patched versions of llvm-project and flang-llvm.

I tried testing it on grace2, but the cloning of llvm-project appears to freeze grace2 and prevent new ssh sessions.

**TL;DR**: The future of setting up and building OptSched:

```
$ git clone --recursive https://github.com/CSUS-LLVM/OptSched-SuperProject.git
$ cd OptSched-SuperProject && mkdir build && cd build
$ cmake <args> ..
$ ninja
```

-----

Naming is hard. Any suggestions on what to name this? "OptSchedProject" is not that good of a name.

Quincunx271:
Closes #48 

I haven't looked into combining the builds of flang-llvm and llvm proper. The logic gets complicated, because this script allows you to specify that you don't want to build flang and because flang's llvm has a different structure from llvm-project.


Quincunx271:
Any concerns, or can I merge this in?

kerbowa:
I think we need some kind of versioning associated with different tags on the LLVM fork. This makes the default build LLVM 6.0. That may make sense for the project, but it should at least be clear that we are building a really old version of LLVM. When targeting AMDGPU LLVM 6.0 wont work at all. Since our cmake scripts are likely to grow I think that we should organize some of the functions into a seperate file/files within a cmake directory. We need to set this up as if we are planning to support varying builds and different versions of LLVM. Overall, I think this is a really good addition for the project.

Quincunx271:
Versioning sounds like it may be a good idea. We can add a tag; something like `optsched/llvm-6.x' (I don't know if `/` works in tags).

There's nothing to separate into a separate directory yet, but that's a good point. There's nothing precluding someone from cloning OptSched into their own llvm setup, but it would indeed be possible to allow a more configurable superbuild.

There should be little need for many functions and the like though. I strongly prefer simpler build scripts.

You've given me some ideas for improving the organization of this, though. I'll probably make some changes.

kerbowa:
> There's nothing to separate into a separate directory yet, but that's a good point. There's nothing precluding someone from cloning OptSched into their own llvm setup, but it would indeed be possible to allow a more configurable superbuild.

This is what I do since I want to be able to do testing with different versions of LLVM including internal downstream repos. The average LLVM dev will usually be primarily interested in doing testing with a bleeding edge build of LLVM since they will want to know how the scheduler interacts with the latest optimizations.

As long as its easy to extend things so that all the permutations work it sounds good to me.

Quincunx271:
It should now be possible to do different configurations. I also pulled out the flang logic into its own file for organizational purposes.

PR #55, opened 2020-03-27
Fix undefined reference to llvm symbols

Quincunx271:
My recent changes to add unit tests broke the compiler. I didn't notice it because the undefined reference errors only came at link time for the shared library, i.e. at runtime. This fixes that issue, and it also automates the command-line test so that it's harder to break the build in the future.


kerbowa:
I'll give your new testing infrastructure a try soon. I will probably add some IR tests.

Quincunx271:
Just to put this somewhere before it makes its way into the BUILD.md documentation:

From the superbuild build directory, `ninja llvm` builds LLVM and OptSched. `ctest . --output-on-failure` (or any other flags for running ctest) will run the tests. In fact, `ctest . --output-on-failure` will also build OptSched, since one of the tests runs the `check-optsched` target, which depends on OptSched.

PR #56, opened 2020-03-28
Privatize / reduce scope of member variables

Quincunx271:
See #47 

The goal here is to move member variables from `protected` to `private`, or even to remove them altogether in favor of local variables. Reducing the scope of the variable makes the code easier to follow, as it's clearer where the variable is changed.

This is a draft. There's much more work to do. However, I would like to know your thoughts on what's being done before the whole thing is finished.

### DIFF
--- a/include/opt-sched/Scheduler/data_dep.h
+++ b/include/opt-sched/Scheduler/data_dep.h
@@ -389,7 +389,6 @@ protected:
                                 int nodeID, InstCount fileSchedOrder,
                                 InstCount fileSchedCycle, InstCount fileLB,
                                 InstCount fileUB, int blkNum);
-  void AddNode_(SchedInstruction *instPtr, InstCount instNum);
   FUNC_RESULT FinishNode_(InstCount nodeNum, InstCount edgeCnt = -1);
   void CreateEdge_(InstCount frmInstNum, InstCount toInstNum, int ltncy,
                    DependenceType depType);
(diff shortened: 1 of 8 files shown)


Quincunx271:
Marking this as ready for review. There are a lot more classes to do this work on, but I think it's best to keep it to a reviewable chunk at a time.

include/opt-sched/Scheduler/data_dep.h
```diff
@@ -389,7 +389,6 @@ class DataDepGraph : public llvm::opt_sched::OptSchedDDGWrapperBase,
                                 int nodeID, InstCount fileSchedOrder,
                                 InstCount fileSchedCycle, InstCount fileLB,
                                 InstCount fileUB, int blkNum);
-  void AddNode_(SchedInstruction *instPtr, InstCount instNum);
```
vangthao95, line 392: Why some of these methods are being removed like AddNode_() and SetNum()?

Quincunx271: They are not used anywhere. I removed them because there's no point in having them if they aren't moved. They also reduce the ability to modify some of the variables (having a getter and a setter is barely any better than an exposed variable).

If we want them in the future, it's in the git history; we can resurrect the code.

Quincunx271:
There shouldn't be any regressions in performance, since it's all inlinable. I'll run a test to verify this.

Quincunx271:
@vangthao95 I can confirm that there are no regressions in compile-times (times.dat). Thereis a minor change in the number of blocks solved optimally: 7 fewer for fp. I suspect that's in the noise.

vangthao95:
Sounds good to me then.

PR #60, opened 2020-03-30
Add notes on static superiority lecture

Quincunx271:
I figure that it makes more sense to document lectures in the repository than in the GitHub Wiki pages.

### DIFF
--- a/documentation/lectures/GraphTransformations.md
+++ b/documentation/lectures/GraphTransformations.md
@@ -1,0 +1,96 @@
+# Lecture notes on Static Superiority graph transformations
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


kerbowa:
Are you sure you want to detail this publicly if it's unpublished work?

Quincunx271:
I didn't see the problem with that.... Is that something we want to avoid?

vangthao95:
I think we should get the permission of Dr. Shobaki first to post this publicly.

Quincunx271:
It's a bit strange to me, as it's already posted in the code and commented in the code, but okay.

PR #65, opened 2020-04-05
GT: Implement live-range covering (lengthened covered by shortened), plus relaxations

Quincunx271:
This PR implements live-range covering for the static superiority graph transformations, and it also relaxes the rules to apply graph transformations.

The relaxations:

 - Recursive successor requirement:
   - Old: if B uses R which is not used by A, then R is used by C which is recursive successor of B.
   - New: C is a recursive successor of A. (remember that RcrsvScsr(B) is a subset of RcrsvScsr(A)).
     This is valid because the point is to prove that R is used after the "critical" region, and A is at the end of said region.
 - Covering requirements:
   - We can cover with defs, not just with uses. I implemented this by instead counting the net number of registers whose live ranges were lengthened, subtracting off those whose live ranges were shortened. This yields simpler code overall: 4 vectors became 1.

----

This has been tested on the fp part of CPU2006. Without history domination and with PERP, there are no mismatches.

### DIFF
--- a/lib/Scheduler/graph_trans.cpp
+++ b/lib/Scheduler/graph_trans.cpp
@@ -50,19 +50,20 @@ static llvm::SmallVector<const Register *, 10> possiblyLengthenedIfAfterOther(
   return result;
 }
 
-// Converts a class type and a function signature into a pointer to member
-// function.
-// Usage: memfn_t<Cls, int(double) const> ==> int (Cls::*)(double) const
-template <typename Cls, typename FnSig> using memfn_t = FnSig Cls::*;
-
-// Gets the Uses or Defs for the given SchedInstruction.
-template <std::size_t N,
-          memfn_t<SchedInstruction, int16_t(Register **&)> UsesOrDefsFn>
-static llvm::SmallVector<const Register *, N> getSet(SchedInstruction *node) {
+// Gets the Uses for the given SchedInstruction.
+static llvm::ArrayRef<const Register *> getUses(SchedInstruction *node) {
   Register **uses;
-  const int useCount = (node->*UsesOrDefsFn)(uses);
-  // Call the (Iter, Iter) constructor.
-  return {uses, uses + useCount};
+  const int useCount = node->GetUses(uses);
+  assert(useCount >= 0);
+  return {uses, static_cast<size_t>(useCount)};
+}
+
+// Gets the Defs for the given SchedInstruction.
+static llvm::ArrayRef<const Register *> getDefs(SchedInstruction *node) {
+  Register **defs;
+  const int defCount = node->GetDefs(defs);
+  assert(defCount >= 0);
+  return {defs, static_cast<size_t>(defCount)};
 }
 
 GraphTrans::GraphTrans(DataDepGraph *dataDepGraph) {
@@ -241,15 +242,11 @@ bool StaticNodeSupTrans::NodeIsSuperior_(SchedInstruction *nodeA,
   // registers.
   const int regTypes = graph->GetRegTypeCnt();
 
-  const llvm::SmallVector<const Register *, 10> usesA =
-      ::getSet<10, &SchedInstruction::GetUses>(nodeA);
-  const llvm::SmallVector<const Register *, 10> usesB =
-      ::getSet<10, &SchedInstruction::GetUses>(nodeB);
+  const llvm::ArrayRef<const Register *> usesA = ::getUses(nodeA);
+  const llvm::ArrayRef<const Register *> usesB = ::getUses(nodeB);
 
-  const llvm::SmallVector<const Register *, 10> defsA =
-      ::getSet<10, &SchedInstruction::GetDefs>(nodeA);
-  const llvm::SmallVector<const Register *, 10> defsB =
-      ::getSet<10, &SchedInstruction::GetDefs>(nodeB);
+  const llvm::ArrayRef<const Register *> defsA = ::getDefs(nodeA);
+  const llvm::ArrayRef<const Register *> defsB = ::getDefs(nodeB);
 
   // (# lengthened registers) - (# shortened registers)
   // from scheduling B after A. Indexed by register type.
--- a/lib/Scheduler/graph_trans.cpp
+++ b/lib/Scheduler/graph_trans.cpp
@@ -2,8 +2,11 @@
 #include "opt-sched/Scheduler/bit_vector.h"
 #include "opt-sched/Scheduler/logger.h"
 #include "opt-sched/Scheduler/register.h"
+#include "llvm/ADT/ArrayRef.h"
 #include "llvm/ADT/STLExtras.h"
+#include <algorithm>
 #include <list>
+#include <vector>
 
 using namespace llvm::opt_sched;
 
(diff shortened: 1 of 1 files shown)


lib/Scheduler/graph_trans.cpp
```diff
@@ -2,11 +2,71 @@
...
+#include <algorithm>
+#include <vector>
 #include <list>
 
 using namespace llvm::opt_sched;
 
+namespace {
+// Gets the registers which may be lengthened by scheduling nodeB after nodeA,
+// as compared to if nodeB was scheduled before nodeA
+llvm::SmallVector<Register const *, 10> possiblyLengthenedIfAfterOther(
```
kerbowa, line 17: Should be consistent with where const goes in ptr to const types. 
`const Register *`

lib/Scheduler/graph_trans.cpp
```diff
@@ -2,11 +2,71 @@
...
+    SchedInstruction const *nodeB, llvm::ArrayRef<const Register *> bUses,
+    SchedInstruction const *nodeA, llvm::ArrayRef<const Register *> aUses) {
+  llvm::SmallVector<Register const *, 10> result;
```
kerbowa, line 20: Ditto

lib/Scheduler/graph_trans.cpp
```diff
@@ -2,11 +2,71 @@
...
+#include "llvm/ADT/ArrayRef.h"
 #include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/SmallVector.h"
+#include <algorithm>
+#include <vector>
 #include <list>
 
 using namespace llvm::opt_sched;
 
+namespace {
```
kerbowa, line 14: Can these just be static instead of inside an autonomous namespace.

lib/Scheduler/graph_trans.cpp
```diff
@@ -2,11 +2,71 @@
...
+template <typename Cls, typename FnSig> using memfn_t = FnSig Cls::*;
+
+// Gets the Uses or Defs for the given SchedInstruction.
+template <std::size_t N,
+          memfn_t<SchedInstruction, int16_t(Register **&)> UsesOrDefsFn>
+llvm::SmallVector<Register const *, N> getSet(SchedInstruction *node) {
+  Register **uses;
+  const int useCount = (node->*UsesOrDefsFn)(uses);
+  // Call the (Iter, Iter) constructor.
+  return {uses, uses + useCount};
```
kerbowa, line 66: It seems unfortunate we need to constructing vector here for every node's uses. Can we change the underlying type to be a vector, or is that a separate patch?

lib/Scheduler/graph_trans.cpp
Quincunx271: Good point. I'm going to constantly forget this; I do east const most of the time. I'll change this, and I'll see if I can find a tool for myself to catch these errors automatically

lib/Scheduler/graph_trans.cpp
Quincunx271: Somehow, I thought the code was previously producing a vector, but apparently not. I think I'll switch this to an `llvm::ArrayRef` to avoid the copy for now.

Yes, changing the underlying type of the use/def to be a vector is on my radar, but it's probably going to be a while before we get there.

lib/Scheduler/graph_trans.cpp
Quincunx271: Yes, they can. But why do we prefer static functions to anonymous namespace functions?

Quincunx271:
Perhaps this last commit should be its own PR

lib/Scheduler/graph_trans.cpp
kerbowa: It's just a stylistic preference. https://llvm.org/docs/CodingStandards.html#anonymous-namespaces
I agree types are the obvious exception.

kerbowa:
> Perhaps this last commit should be its own PR

Yes that would make it easier to review. They would definitely ask for that on reviews.llvm.org

include/opt-sched/Scheduler/sched_basic_data.h
```diff
@@ -405,15 +404,25 @@ class SchedInstruction : public GraphNode {
...
-  int16_t GetUses(Register **&uses);
-
-  int16_t GetDefCnt() { return defCnt_; }
-  int16_t GetUseCnt() { return useCnt_; }
+
+  int16_t NumDefs() { return defCnt_; }
+  int16_t NumUses() { return useCnt_; }
+
+  llvm::ArrayRef<Register *> GetDefs() {
+    return {defs_, static_cast<size_t>(defCnt_)};
```
kerbowa, line 412: This has the same issue as before right? It just extends it to other parts of the code. I thought we could make the underlying datastrcuture (defs_,uses_) a vector. Is this a lot of work to implement? I know it's in list from #47.

Otherwise I can just review the graphtrans portion of the change without this latest commit.

Quincunx271: I am planning to pull this commit out into its own PR; I just haven't gotten around to that yet.

What do you mean by "same issue as before"? This doesn't copy the underlying datastructure at all. It's just a pointer + size. This also allows the underlying datastructure to be converted into a `vector<Register *>` without any change for users. I did not want to implement conversion to a vector yet because it's tricky to ensure the change doesn't break anything.

kerbowa: Okay this makes sense to me.

lib/Scheduler/graph_trans.cpp
Quincunx271: Thanks, I didn't know it was in the Coding Standards.

Quincunx271:
I pulled that commit into its own branch (https://github.com/Quincunx271/OptSched/tree/refactor-getusedef-arrayref). I'll PR that once this PR gets merged

kerbowa:
Sorry, I'll review this as I get a chance.

lib/Scheduler/graph_trans.cpp
```diff
@@ -2,11 +2,70 @@
...
+
+  return result;
+}
+
+// Gets the Uses for the given SchedInstruction.
+static llvm::ArrayRef<const Register *> getUses(SchedInstruction *node) {
+  Register **uses;
+  const int useCount = node->GetUses(uses);
+  assert(useCount >= 0);
+  return {uses, static_cast<size_t>(useCount)};
```
kerbowa, line 58: Just wondering, does makeArrayRef work here?

Quincunx271: Indeed it would. It calls the same constructor that I'm calling here.

Quincunx271:
@kerbowa I want to make sure I understand what you're saying.

If we have a live-out vreg `R`, meaning that `Exit` uses `R` and thus should be its last user, if we also have an instruction `B` which uses `R` and is the last such instruction before `Exit`?

`R in (Uses(B) intersect Uses(Exit))` and B is the last use of `R` except for `Exit`.

I don't see how that can affect things for this graph transformation. We'll always find `Exit` as a node `C` which also uses `R`.

kerbowa:
> I don't see how that can affect things for this graph transformation. We'll always find `Exit` as a node `C` which also uses `R`.

That's what I mean. I was wondering if live-out values might need special consideration, and should "affect things". For example what if the live range is split and we are spilling the second half of the interval. Could it be better to schedule B first in that situation? This most likely falls out of the realm of what the scheduler should be expected to know when calculating register pressure. Maybe it's worth testing with and without considering the exit node as a valid `C`.

PR #67, opened 2020-04-13
Amdgpu superbuild

Quincunx271:
This adds a superbuild script for the AMD GPU (Optimizer-amd) machine.

There is a problem which I have been unable to work out; perhaps someone has an idea what's causing it: After configuring, you have to repeatedly run `make -j32`, ignoring the errors, until it succeeds. The `.inc` files generated by tablegen don't seem to have their dependencies set up correctly, but I have been unable to debug why this is the case.


Quincunx271:
These latest commits fix the issue. This now works for building on the AMDGPU.

vangthao95:
Is this ready to be merged in? If it is then it looks good to me.

Quincunx271:
Now it should be ready to merge. I'd need to write a descriptive commit message when I squash, but it should be functionally ready.

Quincunx271:
I found another kink to work out: sometimes when OptSched files are changed, `make` doesn't rebuild. Instead, you have to navigate to the ROCm-build directory and kick off `make` yourself.

Quincunx271:
Closing this for now. I see this as superseded by #86

Issue #61, opened 2020-04-02
runspec-wrapper-* needs to update "The list schedule" -> "The initial schedule"; And avoiding this class of bug in the future


Quincunx271:
[This commit](https://github.com/CSUS-LLVM/OptSched/commit/efb291911d3dcfd522a9c61e7135a97acb7e9515#diff-4681df460adddcca76f6d47ad0e7bd65R395) introduced a bug in the runspec-wrapper-* scripts: those scripts look for the regex `The list schedule .* is optimal`, but this commit changes the log message to `The initial schedule ... is optimal`.

To combat future bugs of this kind, I propose we come up with a convention to mark a log message as something we are grepping for in our scripts. For example, this one could be:

```
Logger::Info("<The list schedule> of <length %d> and <cost %d> <is optimal>", ...)
```

That way, we can tell at a glance if the log statement is not just for the human reader, but also for script readers, and what pieces of it must be unchanged (or if they are changed, every relevant script should be updated)

Quincunx271:
Another idea: log JSON, but have the interface to log the event be something like:

```
Logger::Event("EventID", "name", dagName, "length", dagLength, "cost", dagCost);
```

Producing a log output of:

```
EVENT: {"id": "EventID", "name": "Function:Block", "length": 20, "cost": 30}
```

Then, the Python scripts can parse these logs using the standard library module `json`.

kerbowa:
Either sounds good. You could add an option to write the json output to some separate file. That would make it easier to save only the data we care about from a run.

They are good ideas but at least the scripts you guys are currently using would need to be updated to go along.

Quincunx271:
I have changes which use Logger::Event(...) for all the relevant log statements I could find. However, what remains is changing the scripts to parse the new format instead of the old format. I haven't finished doing that yet.

PR #69, opened 2020-04-14
[NFC] Add comments to relevant logs indicating it is used for scripts and fixed a typo.

vangthao95:
Add comments documenting whether a log is used in a script and the relevant files. This tries to address #61. 

The relevant files have not yet been added to the comments. I plan on refactoring the scripts so that common regular expressions and functions are in one file to eliminate redundancies.


Quincunx271:
I have a `Logger::Event(...)` function which spits out json; that will greatly simplify the parsing code and eliminate the need for regular expressions. If we want to go that route, there's no reason to spend time refactoring out common regular expressions if we're just going to remove them.

vangthao95:
Now that I think about it some more, json sounds like a good idea. Let's go that route.

PR #70, opened 2020-04-16
[Enumerator] Fix enumerator not properly updating priorities for dynamic heuristics.

vangthao95:
The enumerator was not properly updating the priorities for instructions already in the ready list for dynamic heuristics like LUC. New instructions added to the ready list would have the proper priority but their priority does not get updated. This change updates the priorities for all instructions in the ready list when a dynamic heuristic is selected.

Regarding compile time statistics for OptSched, FFT with heuristic `LUC_NID` gained +2 optimal regions found in the second pass. PlaidML saw a -1 to -3 loss in the optimal regions found and +1 to 3 gain in timed out and not improved when using LUC as a first priority heuristic. I believe this is a reasonable tradeoff for the new changes.

No significant regressions were seen in execution-time performance. PlaidML reported a -0.26% to 0.15% difference and SHOC reported a 0.09% to 0.60% difference in geometric scores. Both were within random variation.

### DIFF
--- a/include/opt-sched/Scheduler/enumerator.h
+++ b/include/opt-sched/Scheduler/enumerator.h
@@ -2,7 +2,7 @@
 Description:  Defines an enumerator class.
 Author:       Ghassan Shobaki
 Created:      Jun. 2002
-Last Update:  Jun. 2017
+Last Update:  Apr. 2020
 *******************************************************************************/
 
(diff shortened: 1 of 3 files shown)


vangthao95:
I will do a squash & merge and copy all of the details that I summarized here into that squash & merge commit. 

Regarding keyedEntries_, I'm not sure if I want to change that here.

PR #71, opened 2020-04-16
[NFC]Print reverting scheduling message with logger to allow printing in release

VladM1076:
Reverting scheduling statement will now appear in release build.
Vang and I believe this statement is important to print in release to provide a more detailed scheduler output without needing to run the slower debug build.

### DIFF
--- a/example/optsched-cfg/sched.ini
+++ b/example/optsched-cfg/sched.ini
@@ -12,7 +12,7 @@ PRINT_SPILL_COUNTS YES
 # First pass minimizes RP and second pass tries to balances RP and ILP.
 # YES
 # NO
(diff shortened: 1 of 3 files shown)


vangthao95:
I believe this is an important log to print because it basically means the scheduling has failed for the particular region and we must change how we handle it in our compile-time statistics.

Quincunx271:
As an aside, what is NFC?

VladM1076:
[NFC] -> No Functional Changes

vangthao95:
It is part of the LLVM lexicon: https://github.com/llvm/llvm-project/blob/master/llvm/docs/Lexicon.rst

PR #72, opened 2020-04-17
Log events as json

Quincunx271:
Currently putting this as a draft PR because it's not complete; I still need to finish going through all the scripts.

I've run into some problems.

 - runspec-wrapper-optsched.py
   - What does `calculatePeakPressureStats` do? I'm unable to find the log command matching `BLOCK_PEAK_REG_BLOCK_NAME` ( "LLVM max pressure after scheduling for BB" ), nor `BLOCK_PEAK_REG_PRESSURE_REGEX`.
   - Which time is `getBenchmarkResult` grabbing? Does it intend to grab the amount of time scheduling took?
      - It seems that it is getting results for blocks which did not go through OptSched, for example:

        ```
        *************************************
        Function: MAIN_finalize
        GREEDY RA: Number of spilled live ranges: 0

        Stores: 0 Reloads: 0 Spill Count: 0
        Stores without cleanup: 0 Reloads without cleanup: 0 Spill Count without cleanup: 0
        Store Cost: 0.000000e+00 Load Cost: 0.000000e+00 Spill Cost: 0

        SC in Function MAIN_finalize 0
        *************************************
        ```

        I actually don't quite understand why we have logs like this. When does this happen instead of OptSched? Is this just for blocks in functions which are not hot (if HOT_ONLY)?

I'll probably have more questions, but this is where I'm at now

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /build
 /.vscode
+__pycache__
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /build
 /.vscode
 __pycache__
+*.pyc
--- a/include/opt-sched/Scheduler/logger.h
+++ b/include/opt-sched/Scheduler/logger.h
@@ -10,7 +10,12 @@ Last Update:  Mar. 2011
 #define OPTSCHED_GENERIC_LOGGER_H
 
 #include "opt-sched/Scheduler/defines.h"
-#include <iostream>
+#include <array>
+#include <cstddef>
+#include <cstdint>
+#include <iosfwd>
+#include <type_traits>
+#include <utility>
 
 namespace llvm {
 namespace opt_sched {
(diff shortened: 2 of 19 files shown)


vangthao95:
`BLOCK_PEAK_REG_BLOCK_NAME` and `BLOCK_PEAK_REG_PRESSURE_REGEX` aren't used anymore and the logging related to them has been removed in commit 7aecbc5a0b05035d6086141446b93feeb9e665a3 back in October of 2018.

>Which time is `getBenchmarkResult` grabbing? Does it intend to grab the amount of time scheduling took?
Yes, it runs each benchmark then records the total time for that benchmark and not just the B&B enumeration time.

```
*************************************
Function: MAIN_finalize
GREEDY RA: Number of spilled live ranges: 0

Stores: 0 Reloads: 0 Spill Count: 0
Stores without cleanup: 0 Reloads without cleanup: 0 Spill Count without cleanup: 0
Store Cost: 0.000000e+00 Load Cost: 0.000000e+00 Spill Cost: 0

SC in Function MAIN_finalize 0
*************************************
```
I believe this is used to estimate the actual spills (not sure if it's the actual or the estimate) and is applied by the patch https://github.com/CSUS-LLVM/OptSched/blob/master/patches/llvm6.0/llvm6-print-spilling-info.patch. This is done during a different pass in LLVM and not the scheduler so it won't be possible to print this information with OptSched.

Quincunx271:
Would it be beneficial to log both this style and the INFO: regex-parsing style? That way, we wouldn't have to change _everything_ at once, but we could make the changes to the scripts over time. Once all the regexes have been changed, _then_ we could remove the INFO logs.

vangthao95:
We can log both styles for now and gradually update the scripts that we use. I am fine with either though. We only use a few scripts so it wouldn't take long to update them. I don't think we should bother with old scripts we don't use anymore.

Quincunx271:
@vangthao95 Do you know which few scripts we are using?

vangthao95:
I believe we are only using the scripts `CPU2006/runspec-wrapper-optsched.py`, `misc/validation-test.py`, and the plaidbench scripts.

Quincunx271:
I've cleaned this up. This closes #61 .

vangthao95:
LGTM.

Quincunx271:
I believe the state of this PR was: I need to run larger scale tests.

Quincunx271:
I'm in the process of running larger scale tests for this. Please let me know what you think about the JSON format and its readability:

Here's a snippet of the logs in Gems (I removed the duplicated INFO logs for this readability verification):

```
EVENT: {"event_id": "ProcessDag", "name": "MAIN_:2", "num_instructions": 7, "max_latency": 0, "time": 0}
EVENT: {"event_id": "HeuristicResult", "length": 7, "spill_cost": 0, "cost": 0, "time": 0}
EVENT: {"event_id": "CostLowerBound", "cost": 700, "time": 0}
EVENT: {"event_id": "HeuristicScheduleOptimal", "length": 7, "cost": 0, "time": 0}
EVENT: {"event_id": "BestResult", "name": "MAIN_:2", "cost": 0, "length": 7, "optimal": true, "time": 0}
EVENT: {"event_id": "BestLocalRegAllocSimulation", "dag_name": "MAIN_:2", "num_spills": 109, "num_stores": 109, "num_loads": 0, "time": 0}
EVENT: {"event_id": "LocalRegAllocSimulationChoice", "dag_name": "MAIN_:2", "num_spills": 109, "num_stores": 109, "num_loads": 0, "time": 0}
EVENT: {"event_id": "ScheduleVerifiedSuccessfully", "time": 0}
```

Here's the corresponding logs before this change:

```
INFO: Processing DAG MAIN_:2 with 7 insts and max latency 0. (Time = 0 ms)
INFO: The list schedule is of length 7 and spill cost 0. Tot cost = 0 (Time = 0 ms)
INFO: Lower bound of cost before scheduling: 700 (Time = 0 ms)
INFO: The initial schedule of length 7 and cost 0 is optimal. (Time = 0 ms)
INFO: Best schedule for DAG MAIN_:2 has cost 0 and length 7. The schedule is optimal (Time = 0 ms)
INFO: OPT_SCHED LOCAL RA: DAG Name: MAIN_:2 Number of spills: 109 (Time = 0 ms)
INFO: Number of stores 109 (Time = 0 ms)
INFO: Number of loads 0 (Time = 0 ms)
INFO: Schedule verified successfully (Time = 0 ms)
```

Quincunx271:
Future idea: gather these events into a relational database to make it easier to answer questions about the effects of things.

Quincunx271:
A report of possible differences (ran on my own machine while multitasking with other things):

Compile time increase: 1292s -> 1378s
Spills: no difference
Blocks: Mild degradation:
 - +2 blocks non optimal and not improved. Of which:
   - -1 block optimal and improved
   - -1 block optimal but not improved

I did only one run. The only Logger::Event calls within the B&B enumeration is "Enumerating" at a given target length. As such, I think this cost should be okay, although I'm thinking of profiling `Logger::Event` to ensure it doesn't significantly reduce performance.

Quincunx271:
The `Logger::Event` function call is slower than a corresponding `Logger::Info` function call by about 10-30%, but that difference can be removed with some micro optimization. The duplicated `Logger::Info` call after the `Logger::Event` call in many locations will be making a bigger difference than any micro optimization.

Issue #57, opened 2020-03-28
History Domination: bug with PERP


Quincunx271:
When running cactus with PERP and APPLY_HISTORY_DOMINATION set to YES, the `ApplyBndRobin:63` block has a spill cost of 4, when it should be 2.

sched.ini file follows. I set `ApplyBndRobin` as the only hot function so that it wouldn't waste time compiling the rest of the benchmark:

```
# Use optimizing scheduling
# YES
# NO : No scheduling is done.
# HOT_ONLY: Only use scheduler with hot functions.
USE_OPT_SCHED HOT_ONLY

# Print spill counts
# Same options as use optimal scheduling.
PRINT_SPILL_COUNTS YES

# Use two pass scheduling approach. 
# First pass minimizes RP and second pass tries to balances RP and ILP.
# YES
# NO
USE_TWO_PASS NO

# These 3 flags control which schedulers will be used.
# Each one can be individually toggled. The heuristic
# list scheduler or ACO must be run before the 
# enumerator.
# VALUES:
# YES
# NO
# HEUR_ENABLED is the Heuristic scheduler.
HEUR_ENABLED YES
# ACO_ENABLED is the Ant Colony Optimization scheduler.
ACO_ENABLED NO
# ENUM_ENABLED is the Branch and Bound scheduler.
ENUM_ENABLED YES

# Controls when ACO should be run, either before or
# after the enumerator. Both can be enabled at the
# same time. ACO is disabled if both are disabled.
# Run ACO before the enumerator
# VALUES:
# YES
# NO
ACO_BEFORE_ENUM YES
# Run ACO after the enumerator
ACO_AFTER_ENUM NO

# A time limit for the whole region (basic block) in milliseconds. Defaults to no limit.
# Interpretation depends on the TIMEOUT_PER setting.
# Not used when the two pass scheduling approach is enabled.
REGION_TIMEOUT 5

# A time limit for each schedule length in milliseconds. Defaults to no limit.
# Interpretation depends on the TIMEOUT_PER setting.
# Not used when the two pass scheduling approach is enabled.
LENGTH_TIMEOUT 5

# A time limit for the whole region in milliseconds. Defaults to no limit.
# Only used when two pass scheduling is enabled.
# A time limit for the whole region.
FIRST_PASS_REGION_TIMEOUT 5
# A time limit for each schedule length.
FIRST_PASS_LENGTH_TIMEOUT 5

# A time limit for the second pass in milliseconds.
[...]

PR #74, opened 2020-04-23
[HistDom] Fix RP pruning for peak SCF

kerbowa:
We were incorrectly considering the total cost which is stored in the history
node as the suffix cost. The total cost could also be determined by the
history prefix cost.


vangthao95:
LGTM.

Quincunx271:
I've tested this against the mismatch found in #57 . The mismatch goes away. I'm currently running a larger test over all of fp.

Quincunx271:
My larger test appears to indicate that this fixes the problem.

PR #75, opened 2020-04-26
GitHub workflows

Quincunx271:
I spent a little time here and there setting up GitHub Workflows (under the Actions tab) which run on PR and push. I've got 4 here:

* Ensure that the code compiles and that the unit tests pass.
* Ensure the codebase is clang-format-ed according to clang-format-9 (version is changeable to a different number if we want)
* Ensure the changes don't introduce any east const. This actually checks if a changed file has east const; currently, it seems that only **data_dep.cpp** does.
* Run clang-tidy-9 and ensure there are no warnings. There are currently some warnings. I disabled `readability-identifier-naming`, as there are a very large amount of violations there.

Currently failing: clang-tidy, clang-format.

----

This PR has a lot of commits; that's because I was testing the changes by pushing to master on my fork; each commit triggers a run. I intend to squash this on merge and write up a better commit message.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -1,0 +1,62 @@
+name: C/C++ CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install ninja
+        id: ninja-build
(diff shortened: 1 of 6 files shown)


Quincunx271:
I'm hesitating to merge this because most of the checks fail (2/5 pass). Each failed check sends us an email. I have patches which fix the checks. I'm thinking maybe it would be nicer to merge those fixes before this, so that we don't get spammed by 10s of emails saying that the check failed.

PR #77, opened 2020-04-29
Fix clang-tidy warnings

Quincunx271:
#75 has a check which flags these clang-tidy warnings


Quincunx271:
@kerbowa What does "Jez" mean?

PR #80, opened 2020-04-30
Fix compute critical path bug.

vangthao95:
This patch fixes a crash in debug mode due to a failed [assert](https://github.com/CSUS-LLVM/OptSched/blob/master/lib/Scheduler/sched_basic_data.cpp#L242) when computing the critical path. When computing the critical path, a dir of DIR_FRWD calculates the critical path starting from the root node. To do this, it uses the current critical path of predecessor nodes but we were using the successors instead which returned an invalid value. This was accidentally swapped in #56.

Tested with SPEC CPU2006 and SHOC and this patch fixes the crash in debug mode.

### DIFF
--- a/include/opt-sched/Scheduler/graph.h
+++ b/include/opt-sched/Scheduler/graph.h
@@ -512,7 +512,7 @@ inline UDT_GEDGES GraphNode::GetRcrsvScsrCnt() const {
 }
 
 inline LinkedList<GraphEdge> *GraphNode::GetNghbrLst(DIRECTION dir) {
-  return dir == DIR_FRWRD ? scsrLst_ : prdcsrLst_;
+  return dir == DIR_FRWRD ? prdcsrLst_ : scsrLst_;
 }
 
 inline GraphEdge *GraphNode::GetFrstScsrEdge() {


paul-mchugh:
Forward means the critical path from the current instruction to the end of the scheduling region. Backward is the critical path from the start of the region to the current instruction

Quincunx271:
Yes, but in the context of GraphNode alone, "forward" seems to be "in the direction of the edges" and "backward" seems to be "in the opposite direction of the edges".

vangthao95:
`GetNghbrLst` is actually only used in one function, [CmputCrtclPath_()](https://github.com/CSUS-LLVM/OptSched/blob/master/lib/Scheduler/sched_basic_data.cpp#L218), which is called when calculating the critical paths [here](https://github.com/CSUS-LLVM/OptSched/blob/master/lib/Scheduler/data_dep.cpp#L3078). In this context, GetNghbrLst seems to be the nodes of those already visited which should be the predecessors (dir forward) or successors (dir backward). I agree that calling it GetNghbrLst isn't intuitive. We should change it to something else. However, for now I will merge this in since this causes a critical bug in the scheduler.

PR #83, opened 2020-05-21
Allow PriorityList<T, K>::BoostEntry to lower priority of instructions in ready list

VladM1076:
Adds a missing piece of functionality from PR #70   Without this change PriorityList<T, K>::BoostEntry fails it's assert (`assert(newKey > entry->key);`) when asked to lower priority of an instruction by changes made in PR #70 when using LUC. 
This change removes that assert since we would like it to able to lower priority, and adds the ability for BoostEntry to lower priority of instructions.

### DIFF
--- a/include/opt-sched/Scheduler/lnkd_lst.h
+++ b/include/opt-sched/Scheduler/lnkd_lst.h
@@ -584,8 +584,7 @@ void PriorityList<T, K>::BoostEntry(KeyedEntry<T, K> *entry, K newKey) {
 
   assert(LinkedList<T>::topEntry_ != NULL);
 
-  if (entry->key < newKey) // behave normally
-  {
+  if (entry->key < newKey) { // behave normally
     entry->key = newKey;
 
     // If it is already at the top, or its previous still has a larger key,
@@ -614,8 +613,8 @@ void PriorityList<T, K>::BoostEntry(KeyedEntry<T, K> *entry, K newKey) {
     assert(next != entry->GetNext());
     LinkedList<T>::RmvEntry_(entry, false);
     InsrtEntry_(entry, next);
-  } else // move entry down on priority list
-  {
+    
+  } else { // move entry down on priority list    
     entry->key = newKey;
 
     // if it is at the bottom or next entry still has a smaller key,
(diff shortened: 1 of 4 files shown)


include/opt-sched/Scheduler/lnkd_lst.h
```diff
@@ -573,43 +573,66 @@ inline T *PriorityList<T, K>::GetNxtPriorityElmnt(K &key) {
...
   KeyedEntry<T, K> *crnt;
   KeyedEntry<T, K> *next = entry->GetNext();
   KeyedEntry<T, K> *prev = entry->GetPrev();
 
-  assert(newKey > entry->key);
   assert(LinkedList<T>::topEntry_ != NULL);
 
-  entry->key = newKey;
+  if (entry->key < newKey) // behave normally
+  {
```
Quincunx271, line 588: Can you put the comment after the brace, or on the next line after the `if`? The LLVM code style places the `{` right after the `)` for the `if`, like so: `if (entry->key < newKey) {`, and it's not like that here because of the comment.

include/opt-sched/Scheduler/lnkd_lst.h
```diff
@@ -573,43 +573,66 @@ inline T *PriorityList<T, K>::GetNxtPriorityElmnt(K &key) {
...
+    }
 
-  assert(next != entry->GetNext());
-  LinkedList<T>::RmvEntry_(entry, false);
-  InsrtEntry_(entry, next);
+    assert(next != entry->GetNext());
+    LinkedList<T>::RmvEntry_(entry, false);
+    InsrtEntry_(entry, next);
+  } else // move entry down on priority list
+  {
```
Quincunx271, line 618: Same thing here: comment after brace

include/opt-sched/Scheduler/lnkd_lst.h
VladM1076: Ok, I have updated the style. Thank you for pointing it out, old habits die hard.

vangthao95:
This looks good but it is failing some of our checks.

VladM1076:
Failed checks were caused by a white space. I have removed it.

vangthao95:
LGTM.

PR #84, opened 2020-05-25
Added print spilling patch for flang LLVM7

paul-mchugh:
This is the patch that is needed to compile flang LLVM7 with spilling info printing.

I am using flang LLVM7 (LLVM 7 is the last version old flang supports) for the ACO paper, so I modified the flang LLVM6 patch so I can get spilling info for this version.

### DIFF
--- a/patches/llvm7.0/README.md
+++ b/patches/llvm7.0/README.md
@@ -1,0 +1,10 @@
+[![CSUS](http://www.csus.edu/Brand/assets/Logos/Core/Primary/Stacked/Primary_Stacked_3_Color_wht_hndTN.png)](http://www.csus.edu/)
+
+# OptSched - Optimizing Scheduler
+This directory contains patches for LLVM 7.0. These patches must be applied before building LLVM 7.0 to print spilling information.
+
+## Files
+
+`flang-llvm7-print-spilling-info.patch`
+
+This file is for Flang LLVM 7.0 located under the release_70 branch at https://github.com/flang-compiler/llvm.
--- a/patches/llvm7.0/flang-llvm7-print-spilling-info.patch
+++ b/patches/llvm7.0/flang-llvm7-print-spilling-info.patch
@@ -1,0 +1,216 @@
+From ea1d128a9988714cff12d591f7b1f846335ba5f0 Mon Sep 17 00:00:00 2001
+From: Paul McHugh <paulpjmmchugh@gmail.com>
+Date: Mon, 25 May 2020 11:33:41 -0700
+Subject: [PATCH] Patch for adding spilling recording code(flang/llvm7)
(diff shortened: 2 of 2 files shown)


Quincunx271:
Can you remove the trailing whitespace from the file? (shown at: https://github.com/CSUS-LLVM/OptSched/runs/707157717 ). If you are using vscode, there's a command to trim trailing whitespace. Please make a new commit which fixes this.

paul-mchugh:
I don't think that altering the diff is a good idea.  The trailing white space in question are single spaces inserted by git format-patch in the first character of a line(also the last character if the line is empty). Patch uses the lines surrounding the additions/removals to help apply the patch when the edit positions alone can't apply the patch(eg: someone added/removed stuff from the start of the file).  
Furthermore, the patches are not themselves c++ code, so I don't think it makes sense to make them comply with any particular coding standards.  The patches will never be integrated into another code base, and will (probably) never be edited.  I don't think concerns about leading/trailing spaces are relevant here.

Quincunx271:
I did not know that git format-patch did this. It's worth noting that I made a commit a while back which did strip the whitespace from .patch files: https://github.com/CSUS-LLVM/OptSched/pull/76/files#diff-ebda4214553aa30d2d78cbc2e0903b8a . I just tested it, and those patches work just fine despite the trailing whitespace being removed.

The check actually checks every file in the repo, not just the ones changed by the PR, which is why I want the trailing whitespace removed. With this file committed, every PR will be flagged as having trailing whitespace. I strongly prefer having every file conform to no trailing whitespace, but I should be able to change the script to make it only flag trailing whitespace in modified files, if that's the direction that we want to go.

paul-mchugh:
patch will only try to use the surrounding fragment lines if other merging methods fail.  Can you exclude .patch files from the trailing ws check?

Quincunx271:
The script uses git diff-tree to check whitespaces, which cannot on its own exclude files. I'll look to see how to exclude just the .patch files, or perhaps just the patches directory.

Quincunx271:
If I use git diff instead, I can control precisely which files to check. For my reference when I fix the script tomorrow:

```
git diff --check $(git hash-object -t tree /dev/null) HEAD -- '*.py' '*.cpp' # and so on
```

(https://stackoverflow.com/questions/8554776/how-to-filter-git-diff-based-on-file-extensions for the git diff on specific files only)

I'll also revert the changes to the previously modified patch files when I make this fix.

paul-mchugh:
I think this command will also work.  It only runs the check on files in the tree that are not in the patches directory.
git diff-tree --check $(git hash-object -t tree /dev/null) HEAD $(ls | sed -E '/.*(patches).*/d')

PR #85, opened 2020-05-26
Fix whitespace check to skip .patch files

Quincunx271:
git format-patch produces trailing whitespace in its output. We should
respect that and allow whitespace in those files.

Some .patch files previously had their trailing whitespace removed.
Their whitespace has now been restored.

----

Fixes the issue raised in the comments of #84


paul-mchugh:
LGTM

vangthao95:
LGTM.

PR #86, opened 2020-05-27
Rewrite CMake scripting to support standalone

Quincunx271:
These changes allow us to be built standalone, provided an existing LLVM
build can be found with find_package(LLVM).

This enables us to share a single build of LLVM rather than rebuilding
LLVM for each build configuration. It also makes it possible to use a
different configuration for our code as compared to LLVM, which may aide
debugging.

If unit tests are enabled, the LLVM package must have the
LLVM_INSTALL_UTILS option set so that we can find FileCheck and other
required tools.

-----

This is a draft PR right now for the following reasons:

 - Testing. I still need to test on Grace2 to make sure that things still work even if done in this way.
 - I also need to do some work for the AMDGPU to work. One such case: I have `find_package(LLVM 6.0 ...)`, which will error out for LLVM 7.0 and above.
   - Another AMDGPU case: we use private headers from the LLVM AMDGPU target. I have to provide a different way to get those headers.
 - `extern bool OPTSCHED_gPrintSpills;` needs special care for the standalone build. If the provided LLVM package doesn't have our patch applied, this will cause problems. Instead, I need to somehow detect that case and either error out, or else mark this variable not extern.

Another annoying point is that I had to make the standalone case download llvm-project to get access to GTest and llvm-lit in a way that the LLVM cmake functions understand. This works and doesn't take too long (roughly 10 seconds on my machine), but it's a wart.

----

Thoughts building off of this:

 - We could have a single, shared build of llvm (or one build for each desirable configuration) on our machines (Grace2 and the AMD one) which we all link against. This should save on compilation time and disk space.
 - Our custom llvm-project could add a couple packages with the prebuilt LLVM installs for the configuration. This can be used in our CI to greatly speed it up, although perhaps if we do that, it should be in _addition_ to what we have now, to test both configurations.


lib/Wrapper/OptimizingScheduler.cpp
```diff
@@ -40,7 +40,7 @@
 using namespace llvm::opt_sched;
 
 // hack to print spills
-bool OPTSCHED_gPrintSpills;
+extern bool OPTSCHED_gPrintSpills;
```
Quincunx271, line 43: Actually, this is the one which should be NOT extern. Rather, the declaration in the patch files (and our fork of llvm-project) should be marked extern.

Quincunx271:
This should in theory allow an external build while still supporting any of our other build methods. In a review, most of what I'm looking for is a verification that this works for your build method. Actually reviewing any files could be helpful too.

CMakeLists.txt
```diff
@@ -1,101 +1,103 @@
...
-get_directory_property(hasParent PARENT_DIRECTORY)
-if(NOT hasParent)
-  message(FATAL_ERROR "You cannot build opt-sched as a standalone project. Place opt-sched in the llvm tree under the \"projects\" directory.")
+if(TARGET LLVMCodeGen)
+  set(llvm_subproject TRUE)
+else()
+  set(llvm_subproject FALSE)
+endif()
+
+if(NOT llvm_subproject)
```
VladM1076, line 20:
```
if(NOT llvm_subproject)
  if(OPTSCHED_ENABLE_AMDGPU)
    find_package(LLVM 9.0 REQUIRED CONFIG)
  else()
    find_package(LLVM 6.0 REQUIRED CONFIG)
  endif()
```
Does this mean only LLVM 9.0 or LLVM 6.0 are supported with this build? Me and Paul have been using LLVM 7.0 for the testing with CPU2017.

Quincunx271: Good point. I just thought of that late last night; I'll have to think about how to support LLVM 7.0 or other versions. It should be fairly straightforward for me to fix this, but I'll have to think about it and test it.

VladM1076: OK sounds good. I will test the new build on Monday most likely since I have my work cut out for today. Once LLVM 7 is supported I can also try to merge this build change into my branch and see if it helps with making HIP and CUDA easier to build as well, which would be very nice.

Quincunx271: The best I could think of is require `-DOPTSCHED_LLVM_VERSION=7.0` if you're configuring with LLVM 7 rather than 6 (or 9 with the AMDGPU).

paul-mchugh (approved):
I tested it compiling for flang.  It works.  LGTM

VladM1076 (approved):
Was able to build both as standalone and as a sub-project. LGTM.

Quincunx271:
Thanks for the work, y'all.

PR #87, opened 2020-06-01
Fix setting LLVM_MUTATIONS option to YES does not enable mutations

paul-mchugh:
When the default machine instruction scheduler ScheduleDAGMILive is created LLVM adds a generic mutation see
https://github.com/llvm/llvm-project/blob/release/7.x/llvm/lib/CodeGen/MachineScheduler.cpp#L3193
When the ScheduleDAGOptSched (our subclass of ScheduleDAGMILive) is created no mutations are
added.  Our option in sched.ini LLVM_MUTATIONS is supposed to add the generic mutations to our scheduler.  It does currently only calls the function which applies the mutations that have already been added.  It does not actually add the mutation in question.  This commit fixes the issue.

### DIFF
--- a/lib/Wrapper/OptimizingScheduler.cpp
+++ b/lib/Wrapper/OptimizingScheduler.cpp
@@ -218,8 +218,7 @@ ScheduleDAGOptSched::ScheduleDAGOptSched(
   MM->convertMachineModel(static_cast<ScheduleDAGInstrs &>(*this),
                           RegClassInfo);
 
-  if (EnableMutations)
-    addLLVMMutations();
+  addLLVMMutations();
 }
 
 void ScheduleDAGOptSched::SetupLLVMDag() {
@@ -249,13 +248,17 @@ void ScheduleDAGOptSched::initSchedulers() {
   SchedPasses.push_back(OptSchedBalanced);
 }
 
-// Add the appropriate LLVM mutations. Called if LLVM_MUTATIONS is set
+// Add the appropriate LLVM mutations.
 void ScheduleDAGOptSched::addLLVMMutations() {
-  addMutation(createCopyConstrainDAGMutation(TII, TRI));
-  // README: if you need the x86 mutations uncomment the next line.
-  // addMutation(createX86MacroFusionDAGMutation());
-  // You also need to add the next line somewhere above this function
-  //#include "../../../../../llvm/lib/Target/X86/X86MacroFusion.h"
+
+  // Adds the LLVM mutations if LLVM_MUTATIONS is set
+  if (EnableMutations) {
+    addMutation(createCopyConstrainDAGMutation(TII, TRI));
+    // README: if you need the x86 mutations uncomment the next line.
+    // addMutation(createX86MacroFusionDAGMutation());
+    // You also need to add the next line somewhere above this function
+    //#include "../../../../../llvm/lib/Target/X86/X86MacroFusion.h"
+  }
 }
 
 // schedule called for each basic block
--- a/lib/Wrapper/OptimizingScheduler.cpp
+++ b/lib/Wrapper/OptimizingScheduler.cpp
@@ -238,9 +238,9 @@ void ScheduleDAGOptSched::SetupLLVMDag() {
   // Apply llvm DAG post processing.
   if (EnableMutations) {
     addMutation(createCopyConstrainDAGMutation(TII, TRI));
-    //README: if you need the x86 mutations uncomment the next line.
-    //addMutation(createX86MacroFusionDAGMutation());
-    //You also need to add the next line somewhere above this function
+    // README: if you need the x86 mutations uncomment the next line.
+    // addMutation(createX86MacroFusionDAGMutation());
+    // You also need to add the next line somewhere above this function
     //#include "../../../../../llvm/lib/Target/X86/X86MacroFusion.h"
     Topo.InitDAGTopologicalSorting();
     postprocessDAG();
--- a/lib/Wrapper/OptimizingScheduler.cpp
+++ b/lib/Wrapper/OptimizingScheduler.cpp
@@ -218,7 +218,7 @@ ScheduleDAGOptSched::ScheduleDAGOptSched(
   MM->convertMachineModel(static_cast<ScheduleDAGInstrs &>(*this),
                           RegClassInfo);
 
-  if(EnableMutations)
+  if (EnableMutations)
     addLLVMMutations();
 }
 
@@ -237,7 +237,6 @@ void ScheduleDAGOptSched::SetupLLVMDag() {
 
   // Finalize live-in
   RPTracker.closeTop();
-
 }
 
 // Add the two passes used for the two pass scheduling approach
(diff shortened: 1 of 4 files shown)


lib/Wrapper/OptimizingScheduler.cpp
```diff
@@ -237,6 +237,11 @@ void ScheduleDAGOptSched::SetupLLVMDag() {
 
   // Apply llvm DAG post processing.
   if (EnableMutations) {
+    addMutation(createCopyConstrainDAGMutation(TII, TRI));
```
vangthao95, line 240: I think we should follow how LLVM add mutations to the DAG. https://github.com/llvm/llvm-project/blob/master/llvm/lib/Target/X86/X86TargetMachine.cpp#L350

Create the DAG, add mutations, then return the DAG object pointer. The AMDGPU target also uses `SetupLLVMDag()` and we don't want this mutation added to the AMDGPU target. Doing it the LLVM way would make the mutations target-dependent.

paul-mchugh: Thanks for the reply.  I have checked the code, and you are right the AMD scheduler doesn't use the generic LLVM mutations.  I didn't want to add it to createOptSched, b/c it is called before the constructor, and consequently before the sched.ini options including LLVM_MUTATIONS are loaded.  The constructor is not a good place b/c it is used by the AMDGPU target as well.  initSchedulers() (the place where the AMDGPU target adds its mutations) is also not a place b/c init schedulers is only called in the two pass configuration.
We could also just remove the LLVM_MUTATIONS option and add the generic mutation to createOptSched() or we could add a virtual function that is responsible for adding LLVM mutations if the option is present.  I am open to any of the aforementioned options.

vangthao95: I don't see a problem with adding the mutations before LLVM_MUTATIONS is loaded. They won't get applied until we call postprocessDAG() in SetupLLVMDag() anyways. By then LLVM_MUTATIONS should be loaded.

Either way I'm fine as long as the mutations doesn't get applied to all targets.

kerbowa: ScheduleDAGMILive::schedule() calls postProcessDAG(). There are two steps to implementing the mutations. They are added via addMutation, but are not actually applied until postProcessDAG is called. For this reason I thought it best to avoid calling SetupLLVMDag if we have already called the generic scheduler. I'm not sure postProcessDAG is only called once in all permutations.

vangthao95: Thanks for bringing that up! That is a good point that I have missed. We don't need the call to `SetupLLVMDag()` before the generic scheduler if the generic scheduler itself already calls `postProcessDAG()`.

paul-mchugh: @vangthao95  I was going to check if ScheduleDAGMILive::schedule() if or any other part of the LLVM code base called postProcessDAG(), when @kerbowa posted his last message.

Quincunx271:
For that matter (forgetting to clang-format), should I add a GitHub workflow to have the GitHub bot clang-format for us? I have a working yml file I made a while back in experimentation, triggered when the PR author comments "Do: Reformat" (or any other trigger we want)

paul-mchugh:
I fixed the last commit I fixed the issues that Vang, Austin, and I identified in their comments.  This patch creates a new virtual function addLLVMMutations() in ScheduleDAGOptSched.  It is called in the ScheduleDAGOptSched constructor.  It is overridden by ScheduleDAGOptSchedGCN to add the AMD specific mutations instead of the generic mutations.  In light of Austin's comment that ScheduleDAGMILive calls postProcessDAG() I removed several needless calls to it.  Furthermore, I fixed a bug in the AMDGPU scheduler whereby the AMD mutations would only be added if two pass was enabled(the AMDGPU code which added the mutations used to be in initSchedulers, which is only called if OptSched is enabled and two pass is enabled).

vangthao95:
I think with this change we no longer apply the mutations unless we also call ScheduleDAGMILive::schedule() somewhere. This means we have to use the LLVM heuristic and enable LLVM_MUTATIONS to apply the mutations. For example two pass on the AMDGPU won't apply mutations with only LUC heuristic since it doesn't call ScheduleDAGMILive::schedule() which calls postProcessDAG.

kerbowa:
What do we think about adding a hook to OptSchedTarget where we have callbacks based on the arch? This avoids the added step of commenting out the X86 mutations. BTW, we have previously targeted ARM, AArch64, SystemZ, AMDGPU. It shouldn't be assumed that X86 is some default, and it also shouldn't be assumed that these targets all want copy constrain.

Can we just add a callback to OST and add an X86 OptSchedTarget?

paul-mchugh:
Vang,
Thanks for catching the issue.  The reason I removed the extra calls to setupLLVMDag is because if postprocessDAG is called more than once the mutations will be applied more than once. This results in the schedule that the LLVM heuristic gives being different from the schedule that stock LLVM w/o OptSched would give.  In the latest commit I added calls to "Topo.InitDAGTopologicalSorting();" and "postprocessDAG();" back to SetupLLVMDag(), and only call SetupLLVMDag() if ScheduleDAGMILive::schedule() was not called.  This ensures that the mutations are always applied and the graph is always constructed before SUnits is converted from LLVM format.

Austin,
I agree that creating more target specific versions of the plugin would eventually be a good idea, however I currently am working on re-doing the results for the ACO paper.  The un-commented out code only adds a generic mutation which LLVM has a default behavior of applying to all targets (in the AMDGPU case this is overridden, but on most platforms it is added).  If we were to add target specific schedulers the generic one should still have this mutation.  The comment regarding the X86 mutation serves to draw attention the need to add the x86 mutation if you want to measure the baseline cost of the LLVM schedule on x86. This commit fixes a bug in OptSched.  I think improvements to the code's handling of different targets would be best addressed in a different commit.

vangthao95:
One last issue is that the mutations will always be applied regardless of the option set for LLVM_MUTATIONS. Adding `if (EnableMutations) { }` back in `SetupLLVMDag()` should solve this last issue.

The rest looks good to me.

paul-mchugh:
The `if (EnableMutations)` is unecessary.  `postprocessDag()` applies all the mutations that have been added.  If no mutations have been added it [does nothing.](https://github.com/llvm/llvm-project/blob/release/7.x/llvm/lib/CodeGen/MachineScheduler.cpp#L831)  `Topo.InitDAGTopologicalSorting()` just performs a topological sort of the nodes.  If no mutations/dependencies have been added it does nothing

vangthao95:
What I mean is there is no way to disable mutations for the AMDGPU target.

paul-mchugh:
Previously the llvm mutations on AMDGPU were enabled if two pass was turned on and did not respect the LLVM_MUTATIONS option.  Turning them on only when two pass was enabled was pretty clearly a bug, I didn't know we wanted them to be disabled if LLVM_MUTATIONS was set to NO.  Currently LLVM_MUTATIONS is set to NO in the example sched.ini.  If we are going to the AMDGPU mutations conditioned on LLVM_MUTATIONS then we should probably make LLVM_MUTATIONS ON by default.

vangthao95:
I thought it did respect the LLVM_MUTATIONS option. Although it does add the mutations to be applied regardless of LLVM_MUTATIONS, it does a check for `EnableMutations` before actually applying them with `postprocessDAG()`.

Primarily the tests done on the AMDGPU were done with the two-pass thus I forgot to enable them when not using two-pass so you are right, this is a bug. Although, right now the mutations for the AMDGPU do not work properly with two-pass and B&B so we shouldn't have it enabled by default.

PR #91, opened 2020-06-09
Fix dag mutations

vangthao95:
There was an issue with the sequential heuristic list scheduler and LLVM mutations in the second pass. This fix postpones adding the artificial edges from the mutations until after the sequential heuristic list scheduler is finished.

### DIFF
--- a/lib/Scheduler/data_dep.cpp
+++ b/lib/Scheduler/data_dep.cpp
@@ -899,7 +899,8 @@ void DataDepGraph::CreateEdge(SchedInstruction *frmNode,
 }
 
 void DataDepGraph::CreateEdge_(InstCount frmNodeNum, InstCount toNodeNum,
-                               int ltncy, DependenceType depType) {
+                               int ltncy, DependenceType depType,
+                               bool IsArtificial) {
   GraphEdge *edge;
 
   assert(frmNodeNum < instCnt_);
@@ -928,7 +929,7 @@ void DataDepGraph::CreateEdge_(InstCount frmNodeNum, InstCount toNodeNum,
     Logger::Info("Creating edge from %d to %d of type %d and latency %d",
                  frmNodeNum, toNodeNum, depType, ltncy);
 #endif
-    edge = new GraphEdge(frmNode, toNode, ltncy, depType);
+    edge = new GraphEdge(frmNode, toNode, ltncy, depType, IsArtificial);
 
     frmNode->AddScsr(edge);
     toNode->AddPrdcsr(edge);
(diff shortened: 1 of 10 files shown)


lib/Scheduler/sched_basic_data.cpp
```diff
@@ -350,7 +350,8 @@ SchedInstruction *SchedInstruction::GetNxtPrdcsr(InstCount *scsrNum,
 
 SchedInstruction *SchedInstruction::GetFrstScsr(InstCount *prdcsrNum,
                                                 UDT_GLABEL *ltncy,
-                                                DependenceType *depType) {
+                                                DependenceType *depType,
+                                                bool *IsArtificial) {
```
kerbowa, line 354: Pass by reference?

lib/Scheduler/data_dep.cpp
```diff
@@ -2972,8 +2973,15 @@ bool InstSchedule::VerifyDataDeps_(DataDepGraph *dataDepGraph) {
...
     UDT_GLABEL ltncy;
     DependenceType depType;
-    for (SchedInstruction *scsr = inst->GetFrstScsr(NULL, &ltncy, &depType);
-         scsr != NULL; scsr = inst->GetNxtScsr(NULL, &ltncy, &depType)) {
+    bool IsArtificial;
+    for (SchedInstruction *scsr =
+             inst->GetFrstScsr(NULL, &ltncy, &depType, &IsArtificial);
+         scsr != NULL;
+         scsr = inst->GetNxtScsr(NULL, &ltncy, &depType, &IsArtificial)) {
+      // Artificial nodes are not required for the schedule to be correct
```
kerbowa, line 2981: Artificial edges

vangthao95:
> Is the idea to always ignore artificial edges? I thought we needed to do this conditionally during sequential scheduling, or am I misunderstanding the idea?

We only ignore artificial edges during sequential scheduling and when verifying the schedule for correctness. One issue I was having was that when we ignore artificial edges during sequential scheduling and then verify the schedule with artificial edges, it will deem the schedule as invalid. This happens if the B&B enumerator was unable to find a schedule with artificial edges enabled. Artificial edges are not required for correctness and this was just my quick fix. If there are any issues with this, please let me know.

lib/Scheduler/sched_basic_data.cpp
vangthao95: I copied the way it was done to retrieve an optional value from a GraphEdge. Is there an issue with this? Although I do agree we can definitely use a better method to retrieve these values.

Quincunx271:
FYI, you don't need to manually run clang-format. I set it up so that GitHub will do it for you if you add a comment with exactly the text: `Do: Reformat`

vangthao95:
> FYI, you don't need to manually run clang-format. I set it up so that GitHub will do it for you if you add a comment with exactly the text: `Do: Reformat`

Do I just put it in the description for the commit?

Quincunx271:
No, you just comment on the PR (this comment box) and your commits will be reformatted.

vangthao95:
Gotcha. Will it also update future commits?

Quincunx271:
No. It just rewrites the existing commits. Adding another comment in the future will update any added commits. You can delete the comment after the clang-format run has completed.

lib/Scheduler/sched_region.cpp
```diff
@@ -244,6 +257,10 @@ FUNC_RESULT SchedRegion::FindOptimalSchedule(
 #endif
   }
 
+  // Log the lower bound on the cost, allowing tools reading the log to compare
+  // absolute rather than relative costs.
+  Logger::Info("Lower bound of cost before scheduling: %d", costLwrBound_);
```
Quincunx271, line 262: Is there a reason why this was moved away from the lower bound cost calculation as opposed to being directly after it?

vangthao95: No, I must have accidentally placed it too far when I was making this change. There's should be no reason as to why this is so far below the lower bound calculation itself.

Quincunx271: Okay, good to know.

PR #95, opened 2020-06-11
Enable assertions in GitHub workflow

### LINKED ISSUE #81: GitHub workflow should enable asserts
.github/workflows/ccpp.yml should build with asserts enabled rather than disabled. It currently does not, because a Debug build takes too much memory and CMake's Release configuration passes `-DNDEBUG`. Ideally, we would still have asserts enabled.

---

I don't plan on working on this yet. However, I figured that documenting this concern with an issue would be best.

Quincunx271:
Turns on `assert(...)`s in the CI build.

Closes #81

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -37,6 +37,8 @@ jobs:
              -DLLVM_PARALLEL_LINK_JOBS=1 ../cmake/superbuild
              -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_C_COMPILER_LAUNCHER=ccache
              -DCMAKE_CXX_COMPILER=g++ -DCMAKE_C_COMPILER=gcc
+             -DOPTSCHEDSUPER_LLVM_EXTRA_CMAKE_ARGS='-DLLVM_ENABLE_ASSERTIONS=ON'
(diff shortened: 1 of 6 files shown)


Quincunx271:
I made some changes to the CMakeLists.txt which warrant a re-review

Quincunx271:
I'm adding everyone as a reviewer mainly so that we aren't blindsided by the change in the CMakeLists.txt. I replaced `SET(OPT_SCHED_DEBUG_FLAGS "")` with `set(OPTSCHED_EXTRA_DEFINITIONS "" CACHE STRING "Extra add_definitions to pass to OptSched, ;-separated")`. Not only does the variable have a different name, but it's now a CACHE variable.

What this means:

 - You should `;` separate instead of space separate definitions.
 - You can specify its value on the commandline: `cmake '-DOPTSCHED_EXTRA_DEFINITIONS=-DIS_DEBUG_DEFS_AND_USES;-DIS_DEBUG_DEF_USE_COUNT' .`
 - If you update the value in that `set(...)` command, unless it's the first configure, nothing will happen. Instead, either specify its value on the commandline as above, edit the value in the CMakeCache.txt, or add `FORCE` just before the closing `)` (like so: `set(OPTSCHED_EXTRA_DEFINITIONS "" CACHE STRING "Extra add_definitions to pass to OptSched, ;-separated" FORCE)`).

PR #96, opened 2020-06-12
Use Python rather than Bash for run-plaidbench

Quincunx271:
Our non-trivial scripts are mostly in Python, not Bash. Moving to Python
means that there are fewer languages and quirks that we need to learn.

This also adds a parameter for the output directory, rather than
hardcoding it.


vangthao95:
LGTM.

PR #97, opened 2020-06-23
Fix mismatched new/delete

Quincunx271:
Array new/delete cannot be mismatched with regular new/delete. Fix this
mismatched new/delete by replacing the pointer with llvm::SmallVector.
In the process, some member variables were embedded directly rather than
heap-allocated. This is practically free. At most, sizeof(ReadyList)
increases by the size of one pointer, which is negligible in this class.

------

This is not the cause of the undefined-behavior based mismatch of graph transformations on the GPU. However, valgrind did flag this when I was trying to debug said mismatch.

### DIFF
--- a/lib/Scheduler/ready_list.cpp
+++ b/lib/Scheduler/ready_list.cpp
@@ -127,9 +127,8 @@ void ReadyList::CopyList(ReadyList *otherList) {
   assert(otherList != NULL);
 
   // Copy the ready list and create the array of keyed entries. If a dynamic
-  // heuristic is not used then the second parameter should be a nullptr and the
-  // array will not be created.
-  prirtyLst_.CopyList(&otherList->prirtyLst_, keyedEntries_.data());
+  // heuristic is not used then the second parameter should be an empty array.
+  prirtyLst_.CopyList(&otherList->prirtyLst_, keyedEntries_);
 }
 
 unsigned long ReadyList::CmputKey_(SchedInstruction *inst, bool isUpdate,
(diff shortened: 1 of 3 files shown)


lib/Scheduler/ready_list.cpp
```diff
@@ -120,30 +114,22 @@ ReadyList::ReadyList(DataDepGraph *dataDepGraph, SchedPriorities prirts) {
...
+void ReadyList::CopyList(ReadyList *otherList) {
+  assert(prirtyLst_.GetElmntCnt() == 0);
+  assert(latestSubLst_.GetElmntCnt() == 0);
+  assert(otherList != NULL);
 
   // Copy the ready list and create the array of keyed entries. If a dynamic
   // heuristic is not used then the second parameter should be a nullptr and the
   // array will not be created.
-  prirtyLst_->CopyList(othrLst->prirtyLst_, keyedEntries_);
+  prirtyLst_.CopyList(&otherList->prirtyLst_, keyedEntries_.data());
```
vangthao95, line 132: You might also have to modify [CopyList](https://github.com/CSUS-LLVM/OptSched/blob/master/include/opt-sched/Scheduler/lnkd_lst.h#L181).

Quincunx271: CopyList is getting exactly the same parameters as before. I don't follow what I'd need to modify in CopyList

Quincunx271: Although I just realized something I need to investigate: does `keyedEntries_.data()` return `nullptr` when we need it to?

vangthao95: Sorry I wasn't aware data() returns a pointer so this is good.

Since keyedEntries is now a smallVector, we could just pass the vector to CopyList and check for the size or if it isn't empty.

vangthao95:
LGTM.

PR #101, opened 2020-07-18
Updated hot functions list for FP benchmarks in CPU2006 with LLVM7

vangthao95:
This patch adds another example config directory which contains an updated hot functions list for the FP benchmarks in CPU2006 when using LLVM7. The hot functions for the INT benchmarks were removed since they are outdated.

The hot functions were gathered with `perf record` and I confirmed our scheduler was able to see the functions while scheduling.

### DIFF
--- a/example/llvm7-CPU2006-cfg/hotfuncs.ini
+++ b/example/llvm7-CPU2006-cfg/hotfuncs.ini
@@ -1,0 +1,290 @@
+# A list of functions that we should apply optimizing scheduling to.
+# Only applies when the 'USE_OPT_SCHED' option is set to 'HOT_ONLY'.
+
+# ======================================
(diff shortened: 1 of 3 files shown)


vangthao95:
Do: Reformat

Quincunx271:
Weird. Somehow the github action bots change didn't trigger the checks. There was some trailing whitespace in here.

vangthao95:
I just noticed that too. Can the bot be modified to check for that?

Quincunx271:
It will take a lot of experimentation. Not the highest priority at this point, but I'll experiment some in my own repo and see if I can fix it.

vangthao95:
That's fine if we can't get it to work. We should just take note of this for future commits.

PR #103, opened 2020-07-25
ACO print pheromone table on graph

paul-mchugh:
This pull request adds the debugging/troubleshooting functionality to ACO.  The code prints out Graphviz ".dot" graph files which contain information about the state of the ACO pheromone table, the ants that have crossed it, and the the heuristics they encountered.  I also added a script that conveniently converts the Graphviz ".dot" files to pdfs.


Quincunx271:
FYI, you don't need to manually run clang-format. I set it up so that GitHub will do it for you if you add a comment on this PR with exactly the text: Do: Reformat

paul-mchugh:
Thanks for the information Justin.  I will know to do this in the future.  Since it didn't reformat it automatically I had assumed that you hadn't added the re-formatter yet, or had decided against doing so.

Quincunx271:
Thanks for the feedback on that. If the clang-format check printed out something saying how to do this auto-reformatting when it failed, would you have noticed it?

paul-mchugh:
This is only failing the trailing whitespace check because of whitespace in hotfuncs.ini.  I will re-run the checks once the patch fixing hotfuncs.ini is added.

paul-mchugh:
@vangthao95 Any objections to merging now?

vangthao95:
LGTM.

PR #104, opened 2020-07-28
Fix virtual function in base class's constructor not calling derived class's addLLVMMutations.

vangthao95:
C++ does not allow calling a derived class's virtual function from the base class. Thus when using the AMDGPU target, the incorrect LLVM mutations were being applied and the AMD mutations were never added. See here: https://isocpp.org/wiki/faq/strange-inheritance#calling-virtuals-from-ctors

This patch copies how LLVM adds their mutations to be applied which is directly after creating the scheduler. See here: https://github.com/llvm/llvm-project/blob/release/6.x/llvm/lib/CodeGen/MachineScheduler.cpp#L3195

Note, the mutations won't be applied until we call postprocessDAG() which requires EnableMutations to be set to true or if called by ScheduleDAGMILive::schedule().

Tested on PlaidML (densenet121) and SHOC (FFT, GEMM, MD, Sort, Spmv, Stencil2D).

### DIFF
--- a/lib/Wrapper/OptimizingScheduler.cpp
+++ b/lib/Wrapper/OptimizingScheduler.cpp
@@ -66,7 +66,14 @@ static constexpr const char *DEFAULT_CFGMM_FNAME = "/machine_model.cfg";
 
 // Create OptSched ScheduleDAG.
 static ScheduleDAGInstrs *createOptSched(MachineSchedContext *C) {
-  return new ScheduleDAGOptSched(C, llvm::make_unique<GenericScheduler>(C));
+  ScheduleDAGMILive *DAG =
+      new ScheduleDAGOptSched(C, llvm::make_unique<GenericScheduler>(C));
+  DAG->addMutation(createCopyConstrainDAGMutation(DAG->TII, DAG->TRI));
+  // README: if you need the x86 mutations uncomment the next line.
+  // addMutation(createX86MacroFusionDAGMutation());
+  // You also need to add the next line somewhere above this function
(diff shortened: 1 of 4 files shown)


lib/Wrapper/OptimizingScheduler.cpp
```diff
@@ -237,8 +242,10 @@ void ScheduleDAGOptSched::SetupLLVMDag() {
   // Finalize live-in
   RPTracker.closeTop();
 
-  Topo.InitDAGTopologicalSorting();
-  postprocessDAG();
+  if (EnableMutations) {
+    Topo.InitDAGTopologicalSorting();
+    postprocessDAG();
+  }
```
Quincunx271, line 248: I don't quite understand why these should only be run if mutations are enabled.

vangthao95: I am not exactly too sure about `Topo.InitDAGTopologicalSorting()` myself but I believe the information is used when the mutation adds the edges. `addMutation()` only adds the desired mutations to a list and does not apply it until `postprocessDAG();` is called.

Quincunx271: I'm not sure about `Topo.InitDAGTopologicalSorting()` ( https://llvm.org/doxygen/classllvm_1_1ScheduleDAGTopologicalSort.html#a44c5faa549f250a26b1303eb1a3ebd47 ) either. I don't know if we use the topological sort anywhere. If this has been tested and it works, then I think it should be fine.

PR #111, opened 2020-08-17
Add debugging scripts

### LINKED ISSUE #89: Debugging scripts from Wiki
I have a couple scripts which aide in debugging mismatches. Are they useful enough to pull into the main repo?

 - [findblock.py](https://github.com/CSUS-LLVM/OptSched/wiki/Debugging#minimize-the-test-case) searches for which benchmark in CPU2006 has the block, by searching through a spills.dat.
 - [ddg2dot.py](https://github.com/CSUS-LLVM/OptSched/wiki/Debugging#visualizing-the-ddg) converts the output of our `DataDepGraph::WriteToFile(...)` function into a `.dot` file which Graphviz's `dot` executable can visualize for us.

Quincunx271:
I still need to go through several of the other scripts I've written to see if they're more generally useful.

In terms of review, I think the most important thing is that it's clear what the script does and how to use it.

Closes #89

### DIFF
--- a/util/misc/ddg2dot.py
+++ b/util/misc/ddg2dot.py
@@ -1,0 +1,66 @@
+#!/usr/bin/env python3
+import argparse
+import sys
+import re
+
+parser = argparse.ArgumentParser(description='Convert data_dep WriteToFile format to a .dot file')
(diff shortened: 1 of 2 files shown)


vangthao95:
The script is pretty simple and clear. LGTM.

PR #112, opened 2020-08-18
Fix undefined behavior with llvm_unreachable() and require valid values for some settings instead of defaults.

vangthao95:
We were using `llvm_unreachable()` incorrectly and it was causing undefined behavior. One example is when using an incorrect heuristic in combination with another heuristic such as `NOD_CP`, which should be `NID_CP`, it incorrectly gave `LLVM_CP`. This patch removes it in favor of `Logger::Fatal()` which will output an error message and cause the compiler to crash.

Some default values have also been removed and Logger::Fatal() has been put in place to ensure that a valid value is given. We do not want to automatically default to a set value and would rather instead have the program indicate that our value was incorrect.

### DIFF
--- a/include/opt-sched/Scheduler/logger.h
+++ b/include/opt-sched/Scheduler/logger.h
@@ -48,7 +48,7 @@ void RegisterPeriodicLogger(Milliseconds period, void (*callback)());
 void PeriodicLog();
 
 // Shortcuts for each logging level.
-void Fatal(const char *format_string, ...);
+[[noreturn]] void Fatal(const char *format_string, ...);
 void Error(const char *format_string, ...);
 void Info(const char *format_string, ...);
 void Summary(const char *format_string, ...);
--- a/lib/Scheduler/logger.cpp
+++ b/lib/Scheduler/logger.cpp
@@ -97,6 +97,7 @@ void Logger::Fatal(const char *format_string, ...) {
   char message_buffer[MAX_MSGSIZE];
   VPRINT(message_buffer, format_string);
   Output(Logger::FATAL, true, message_buffer);
+  exit(1);
 }
 
 void Logger::Error(const char *format_string, ...) {
--- a/lib/Wrapper/OptimizingScheduler.cpp
+++ b/lib/Wrapper/OptimizingScheduler.cpp
@@ -612,7 +612,8 @@ bool ScheduleDAGOptSched::isOptSchedEnabled() const {
     return false;
   }
 
-  Logger::Fatal("Unrecognized option for USE_OPT_SCHED setting");
+  Logger::Fatal("Unrecognized option for USE_OPT_SCHED setting: %s",
+                optSchedOption.c_str());
 }
 
 bool ScheduleDAGOptSched::isTwoPassEnabled() const {
@@ -624,7 +625,8 @@ bool ScheduleDAGOptSched::isTwoPassEnabled() const {
   else if (twoPassOption == "NO")
     return false;
 
-  Logger::Fatal("Unrecognized option for USE_TWO_PASS setting");
+  Logger::Fatal("Unrecognized option for USE_TWO_PASS setting: %s",
+                twoPassOption.c_str());
 }
 
 LATENCY_PRECISION ScheduleDAGOptSched::fetchLatencyPrecision() const {
@@ -638,7 +640,8 @@ LATENCY_PRECISION ScheduleDAGOptSched::fetchLatencyPrecision() const {
     return LTP_UNITY;
   }
 
-  Logger::Fatal("Unrecognized option for LATENCY_PRECISION setting");
+  Logger::Fatal("Unrecognized option for LATENCY_PRECISION setting: %s",
+                lpName.c_str());
 }
 
 LB_ALG ScheduleDAGOptSched::parseLowerBoundAlgorithm() const {
@@ -649,7 +652,7 @@ LB_ALG ScheduleDAGOptSched::parseLowerBoundAlgorithm() const {
     return LBA_LC;
   }
 
-  Logger::Fatal("Unrecognized option for LB_ALG setting");
+  Logger::Fatal("Unrecognized option for LB_ALG setting: %s", LBalg.c_str());
 }
 
 // Helper function to find the next substring which is a heuristic name in Str
@@ -668,7 +671,7 @@ static LISTSCHED_HEURISTIC GetNextHeuristicName(const std::string &Str,
       return LSH.HID;
     }
 
-  Logger::Fatal("Unrecognized heuristic used!");
+  Logger::Fatal("Unrecognized heuristic used: %s", Str.c_str());
 }
 
 SchedPriorities ScheduleDAGOptSched::parseHeuristic(const std::string &Str) {
@@ -715,7 +718,8 @@ SPILL_COST_FUNCTION ScheduleDAGOptSched::parseSpillCostFunc() const {
     return SCF_TARGET;
   }
 
-  Logger::Fatal("Unrecognized option for SPILL_COST_FUNCTION setting");
+  Logger::Fatal("Unrecognized option for SPILL_COST_FUNCTION setting: %s",
+                name.c_str());
 }
 
 bool ScheduleDAGOptSched::shouldPrintSpills() const {
@@ -730,7 +734,8 @@ bool ScheduleDAGOptSched::shouldPrintSpills() const {
     return HotFunctions.GetBool(functionName, false);
   }
 
-  Logger::Fatal("Unrecognized option for PRINT_SPILL_COUNTS setting");
+  Logger::Fatal("Unrecognized option for PRINT_SPILL_COUNTS setting: %s",
+                printSpills.c_str());
 }
 
 bool ScheduleDAGOptSched::rpMismatch(InstSchedule *sched) {
--- a/lib/Wrapper/OptimizingScheduler.cpp
+++ b/lib/Wrapper/OptimizingScheduler.cpp
@@ -610,11 +610,9 @@ bool ScheduleDAGOptSched::isOptSchedEnabled() const {
     return HotFunctions.GetBool(functionName, false);
   } else if (optSchedOption == "NO") {
     return false;
-  } else {
-    LLVM_DEBUG(dbgs() << "Invalid value for USE_OPT_SCHED" << optSchedOption
-                      << "Assuming NO.\n");
-    return false;
   }
+
+  Logger::Fatal("Unrecognized option for USE_OPT_SCHED setting");
 }
 
 bool ScheduleDAGOptSched::isTwoPassEnabled() const {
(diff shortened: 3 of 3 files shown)


vangthao95:
Do: Reformat

Quincunx271:
Can you push another commit so that the checks will run (checks don't run after github-actions bot makes a change)?

vangthao95:
```
/home/runner/work/OptSched/OptSched/lib/Wrapper/OptimizingScheduler.cpp:616:1: error: control may reach end of non-void function [clang-diagnostic-return-type,-warnings-as-errors]
Use -header-filter=.* to display errors from all non-system headers. Use -system-headers to display errors from system headers as well.
}
^
7 warnings treated as errors
/home/runner/work/OptSched/OptSched/lib/Wrapper/OptimizingScheduler.cpp:628:1: error: control may reach end of non-void function [clang-diagnostic-return-type,-warnings-as-errors]
}
^
/home/runner/work/OptSched/OptSched/lib/Wrapper/OptimizingScheduler.cpp:642:1: error: control may reach end of non-void function [clang-diagnostic-return-type,-warnings-as-errors]
}
^
/home/runner/work/OptSched/OptSched/lib/Wrapper/OptimizingScheduler.cpp:653:1: error: control may reach end of non-void function [clang-diagnostic-return-type,-warnings-as-errors]
}
^
/home/runner/work/OptSched/OptSched/lib/Wrapper/OptimizingScheduler.cpp:672:1: error: control may reach end of non-void function [clang-diagnostic-return-type,-warnings-as-errors]
}
^
/home/runner/work/OptSched/OptSched/lib/Wrapper/OptimizingScheduler.cpp:719:1: error: control may reach end of non-void function [clang-diagnostic-return-type,-warnings-as-errors]
}
^
/home/runner/work/OptSched/OptSched/lib/Wrapper/OptimizingScheduler.cpp:734:1: error: control may reach end of non-void function [clang-diagnostic-return-type,-warnings-as-errors]
}
^
##[error]Process completed with exit code 7.
```

I think it completed some checks before I called the action bot. It gave me this result, I'm not sure if you can see the results from the link: https://github.com/CSUS-LLVM/OptSched/runs/1000139017

Quincunx271:
The fix to the clang-tidy error is to add `[[noreturn]]` to `Logger::Fatal` like so: `[[noreturn]] void Fatal(const char *format_string, ...);`

vangthao95:
I added `[[noreturn]]` and now it gives me this warning:
```
/home/vang/src/original-rocm/opencl/compiler/llvm/projects/OptSched/lib/Scheduler/logger.cpp:100:1: warning: function declared 'noreturn' should not return [-Winvalid-noreturn]
}
^
```

My Fatal():
`[[noreturn]] void Fatal(const char *format_string, ...);`

Quincunx271:
I see why they give that warning. Adding `exit(1);` to the end of the `Fatal(...)` function would make the warning go away. Alternatively (amusingly), adding a call to `llvm_unreachable(...)` at that point should also make the warning go away.

I'd personally go for the `exit(1);` approach. It may increase code size by a tiny amount in Debug mode, but realistically, it's free.

paul-mchugh:
You can add an llvm_unreachable() after the last line of Logger::Fatal().  That will solve the "function declared 'noreturn' should not return" error.

vangthao95:
Thanks for the help! I just pushed the fix and added a return message back to the user with the invalid values that was detected by our scheduler.

Quincunx271:
I have to disagree about using `llvm_unreachable()`. `llvm_unreachable()` (aka `__builtin_unreachable()` in non-debug mode) is dangerous. It tells the compiler to assume that reaching the function call is undefined behavior. If we accidentally make some change which would allow the code to reach that point, rather than exiting, we'd end up with some difficult-to-debug behavior. The compiler could decide to remove our log statement, for example.

While it is safe at this moment, it's not beneficial to use `llvm_unreachable()` because we don't need the increased performance.

paul-mchugh:
```llvm_unreachable()``` only causes undefined behavior if the statement is actually reached (in release mode), just like it is undefined behavior for a program to reach a assert condition in such a state that it would fail the condition.  Also, the primary purpose of ```llvm_unreachable()``` is unambiguously instruct the compiler that it does not need to worry about this branch, the availability of further optimizations is a fringe benefit.  The argument that ```llvm_unreachable()``` potentially causes undefined behavior and therefore should not be used proves too much.  If that argument were valid then there would be no circumstance where its use would be appropriate.

Ultimately I don't think complying with LLVM Coding Standards is that big of a deal since the entire code base will have to be rewritten if this code is to be merged into LLVM.  Both the Logger and the configuration readers would have to be replaced with the relevant LLVM utilities in their entirety.  I am okay with ignoring the coding standards here and elsewhere if you guys think it is prudent.

**Comment by Quincunx271** [comment]
> just like it is undefined behavior for a program to reach a assert condition in such a state that it would fail the condition.

This is actually not true. It's well defined behavior for the program to reach an assert which would fail. If `NDEBUG` is defined, nothing happens. Otherwise, the program outputs a diagnostic and calls `std::abort()`.

Unlike `assert(...)`, `llvm_unreachable()` turns into `__builtin_unreachable()` in release mode. `__builtin_unreachable()` marks the branch as impossible, allowing the compiler to optimize assuming so. If your code would end up reaching the `__builtin_unreachable()` call, something weird will happen. It's never easy to debug, as we saw in this case, where `ANYTHING_CP` was outputting `LLVM_CP`. Because of this very real possibility of reaching undefined behavior, if you don't need it, don't use `llvm_unreachable()`. This does not go against the LLVM coding standards.

In this case, the call to `Logger::Output()` would have already exited, yes, except it might not if we make changes to `Logger::Output()` in the future, unknowingly introducing this undefined behavior. For that reason, I prefer to not use `llvm_unreachable()` and instead just call `exit(1)`.

**Comment by paul-mchugh** [comment]
What you said about ```assert()``` with NDEBUG is true, the same is true with ```llvm_unreachable()``` . If NDEBUG is set then llvm_unreachable()'s behavior is well defined.  It prints out the message passed as the parameter, and the program crashes.  ```llvm_unreachable()``` is equivalent to ```assert(false)```, except that in debug mode ```llvm_unreachable()``` prints out the message passed to it before terminating.

As an example in RELEASE mode an optimizing compiler could convert the following code:
```
void f(void *ptr)
{
    assert(ptr != NULL); //assert that ptr can never be null here

    //the compiler will see this check as pointless since we are telling it ptr can never be NULL (EVEN IN RELEASE)
    if(ptr != NULL)
        do_thing();
}
```
to:
```void f(void *ptr) {do_thing();}```

Like ```assert(0)``` in debug mode ```llvm_unreachable("My error message here")``` has defined behavior.  In release mode they also both cause branch they are on to be marked as unrechable.  The LLVM standards say to use assert liberally and use ```llvm_unreachable()``` "whenever reaching a particular code path is unconditionally a bug," as is true here.  The reason assert and llvm_unrechable are macros is so that simple compiler implementations that don't understand their special significance will remove them from release code, whereas more sophisticated compilers will use their special meanings for optimizations.

If you want to never use assert or ```llvm_unreachable()``` I am fine with varying from the standards.  I don't wan't to make a mountain out of a mole hill, but this appears to be a situation where the use of ```llvm_unreachable()``` is appropriate.

kerbowa:
If this is an incorrect use of llvm_unreachable then LLVM uses it incorrectly...

Maybe try debugging issues with a debug build.

kerbowa:
I thought we were trying to move closer to LLVM style not trying to justify inventing redundant error checking.

vangthao95:
I think from what Paul said, it would be correct if we do some error handling before `llvm_unreachable` is actually reached. However in the specific case for `GetNextHeuristicName()` it does not do any invalid value handling and allowed the scheduler to reach `llvm_unreachable` causing some undefined behavior.

kerbowa:
llvm_unreachable is error handling.

kerbowa:
If this is common enough of a problem that it's too much to ask people to use debug builds, then use 'llvm ::report_fatal_error'.

vangthao95:
We do use both debug and release builds but this issue pops up unnoticed when using release builds. `llvm_unreachable` does not report any error nor crash to indicate something is wrong when using release builds.

kerbowa:
I'm objecting to is the premise that there is something wrong with llvm_unreachable(), or that we are using it incorrectly here. That is plain wrong. It's like complaining that asserts are not being triggered because you are using a release build. Segfaults and UB is normal if an assert is failing in a debug build and you randomly expect things to work if you optimize out the assert.

> This is actually not true. It's well defined behavior for the program to reach an assert which would fail. If NDEBUG is defined, nothing happens. Otherwise, the program outputs a diagnostic and calls std::abort().

An assert failing often implies that UB could follow.

The argument makes sense if you say that sched.ini options are set dynamically by a user and we need something that will reliably fail. Same is true for command-line options. In that case use 'llvm ::report_fatal_error'.

Quincunx271:
There is nothing wrong with `llvm_unreachable()`. It has its purpose. The changes made in this PR fix incorrect use of `llvm_unreachable()`. It's incorrect because it's easy to reach it via malformed user input.

---

In the case of `Logger::Fatal()`, this is a case where `llvm_unreachable()` is correct. However, I would strongly recommend against it for these reasons:

 - The reason to use `llvm_unreachable()` over printing a message and calling `exit(1)` manually is that it turns into `__builtin_unreachable()` in release mode, allowing the compiler to optimize better by assuming it's undefined behavior to reach the call. We don't need any additional performance in this location. This negates the benefit of `llvm_unreachable()`.
 - While this location _is_ unreachable now, it's only unreachable because of a call to `exit(1)` inside another function which has plenty of reason to change independently of `Logger::Fatal()`. For instance, suppose we add some feature to our logging which is easiest to implement via an early return in `Logger::Output()`. Suddenly, this `llvm_unreachable()` call becomes reached. It's unlikely that this will be caught, because it requires the reviewer to remember `Logger::Fatal()` and this specific `llvm_unreachable()` call and also make the connection that the early return causes us to reach this.
 - Bugs caused by undefined behavior are very difficult to debug. If the `llvm_unreachable()` call becomes reachable, it's conceivable that the compiler could remove our call to log the message altogether and return from `Logger::Fatal()`, resulting in our algorithm continuing in an invalid state. It's a nightmare to debug if something does go wrong here.

There is no real cost to using `exit(1)` over `llvm_unreachable()` here. When programming in C++, you have to be aware of all the ways of causing undefined behavior, so it makes perfect sense to pay peanuts in the coldest of cold code to avoid requiring future readers / changers having to remember about this other way of accidentally invoking undefined behavior.

kerbowa:
> In the case of Logger::Fatal(), this is a case where llvm_unreachable() is correct. However, I would strongly recommend against it for these reasons:

Why would we have an llvm_unreachable() instead of exit()? I was definitely not advocating for that.

I agree with you that since this involves something easily changeable by the user reporting an error makes more sense, which is why I'm suggesting `report_fatal_error`.

With ROCm we do inprocess compilation and applications can install their own signal handlers which we manage. They may be expecting the application to trap to the signal handler if the compiler crashes. In 'report_fatal_error' LLVM does cleanup and then raises SIGABRT. We should probably do the same thing.

vangthao95:
I agree we can definitely use `report_fatal_error` in these cases instead of `Logger::Fatal`. In fact, since `Logger::Fatal` seems to function just as `report_fatal_error`, can we not just move away from `Logger::Fatal` and stick with `report_fatal_error` if we want to adhere to the LLVM standards?

PR #113, opened 2020-08-18
ACO schedule constructor for schedules with stalls

paul-mchugh:
This PR adds the ability for the ACO Scheduler to generate schedules with stalls.  It significantly improves performance when LLVM latencies are enabled and there are instructions with a latency > 1


vangthao95 (approved):
LGTM.

PR #115, opened 2020-08-20
Convert most of our fatal error messages to use LLVM's report_fatal_error

vangthao95:
This patch converts most of our fatal error messages to use `llvm:report_fatal_error` instead of `Logger::Fatal`. Both essentially perform the same functionality but we want to move closer to LLVM style.


Quincunx271:
Is there a reason why the `Logger::Fatal()` calls in buffers.cpp and enumerator.cpp weren't also converted?

vangthao95:
I wasn't sure if `buffers.cpp` was still relevant or not. If it isn't relevant anymore, we should remove it and clean up the codebase.

Regarding `enumerator.cpp`, the chunk of code that the `Logger::Fatal`s are in is still experimental and may be subjected to change anyways.

Issue #116, opened 2020-08-20
Refactor LinkedList?


Quincunx271:
lnkd_list.h is full of custom data structures based around LinkedList. It would be nice if we could scrap these in favor of LLVM data structures, but that may not be realistic. If we do scrap it, we should probably ensure we either don't need our custom features or replace our custom features.

-----

The following is mostly notes to myself:

One particular feature of this LinkedList bit me: the `maxSize` custom allocator. Currently, if you specify a max size, the LinkedList will allocate an array and hand off consecutive elements of the array rather than allocate a new `Entry<T>`. This is good especially because of locality, making it much cheaper to access the elements of the LinkedList compared to if the nodes were scattered around memory (although I'm curious of how big of an effect this has in our code). However, this causes problems with removing elements from the LinkedList, as the only elements which can be removed is the last element in the array.

This causes problems with Dr. Heffernan's ILP node superiority algorithm, which removes edges from the graph. The predecessor list for a GraphNode is a LinkedList with a maxSize. For now, I'm not removing the edges from the graph, but it would be nice to do so.

In this case, the problem can be solved by using a free list of the removed nodes.

PR #118, opened 2020-09-08
Refactor Graph / LinkedList

Quincunx271:
Adds C++ STL style iterators to LinkedList. Adds some accessors and adds a way to remove edges.

### DIFF
--- a/include/opt-sched/Scheduler/graph.h
+++ b/include/opt-sched/Scheduler/graph.h
@@ -142,6 +142,9 @@ public:
   // Returns a pointer to the next successor of the node. Must be called after
   // GetFrstScsr() which starts the successor iterator.
   GraphNode *GetNxtScsr();
+  // Gets an iterable range of the successors of this node
+  const LinkedList<GraphEdge> &GetSuccessors() const;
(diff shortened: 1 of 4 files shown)


Quincunx271:
This PR is ready for review. I do not intend to squash any of the commits, so let me know in the review if you think some should be squashed.

paul-mchugh (approved):
LGTM,

My only note is that I object to the use of linked lists in general, since they are cache inefficient (Including this one that is normally sorted since the size of an entry is 24 bytes, so only 2.66 entries fit in a 64 byte cache line as opposed to 8 pointers that fit in a 64 byte cache line).  Should the use of linked lists prove necessary, then at some time in the future we should replace our Linked List with an std::list.

I realize that now is not an appropriate time to remove all linked lists, or replace them with std::list.

VladM1076 (approved):
LGTM

Quincunx271:
@paul-mchugh I'll link this with #116 just to connect the conversations; I agree with you. I will merge this, but it would be a good idea to remove this custom data structure, and it would be a good idea to remove the linked-list altogether if possible.

PR #122, opened 2020-09-18
Sequential list scheduler fix undefined behavior

VladM1076:
Fix an unaccounted for edge case in SequentialListScheduler::IsSequentialInstruction. When Inst->GetNum() is equal to 0,  crntSched_->GetSchedCycle() is passed a -1. This results in GetSchedCycle reading the value in slotForInst_[-1] which is undefined behavior. If slotForInst_[-1] contains -1 then the SeqListSched is stuck in an infinite loop of scheduling stalls since IsSequentialInstruction never returns true for instruction number 0.

Scheduling instruction number 0 with sequential list scheduler should always be legal since its only predecessor is the OptSched artificial root, which is always trivially legal and has no latency/does not block cycle. So IsSequentialInstruction should return true for when Inst->GetNum() is equal to 0.


lib/Scheduler/list_sched.cpp
```diff
@@ -121,6 +121,11 @@ bool SequentialListScheduler::ChkInstLglty_(SchedInstruction *inst) const {
 
 bool SequentialListScheduler::IsSequentialInstruction(
     const SchedInstruction *Inst) const {
+  // Instr with number 0 is always sequential and legal to schedule
+  // since its predecessor is OptSched artificial root which is always
+  // trivially legal
```
Quincunx271, line 126: I think this is the case. Is it worth adding `assert(...)` to verify this assumption that it will only have one predecessor and its one predecessor is the artificial root? I don't know how difficult it is to add that, though.

Quincunx271 (commented):
The code change LGTM. Please squash the commit history into a single commit with informative message.

lib/Scheduler/list_sched.cpp
VladM1076: I was planning on using "Squash and merge", is that what you are referring to or is there a way to squash commit history before a PR?
As far as the asserts go, checking that instr number 0 has one predecessor is easy with `assert(Inst->GetPrdcsrCnt() == 1);` (Just ran lbm with this assert, not triggered.)
Checking if that predecessor is a root isn't as simple since GraphNode::GetFrstPrdcsr() is not a const function and Inst is a const SchedInstruction * in this function. This means that `assert(Inst->GetFrstPrdcsr()->GetPrdcsrCnt() == 0);` won't compile. I would need to write new code to check if predecessor is root.
From my understanding of SequentialListScheduler I think the asserts are unnecessary since the schedule is always root->0->1->...->artificial leaf with stalls added to account for latencies, which artificial root does not have.
I can still add them if you or other on the team think they will be useful.

Quincunx271: I review commit messages in addition to code, and it's kind of hard to review the message when it's not truly formed until squash-and-merge. I'm fine with just squash-and-merge if you rewrite the commit message as if the final diff is the single commit and don't just keep the default message. It is indeed possible to squash commit history before a PR.

You could use `git rebase -i master`, choosing `squash` (or `s`) as the option for relevant commits (`pick` the first one, `squash` the rest onto it). However, that won't only squash commits, but will instead apply your changes on top of any changes which were made to master since you've started your branch. That will require you to resolve merge conflicts as well.

If you want to deal with merge conflicts separately, you can use `git rebase -i <commit-hash-of-commit-we-branch-from>`, or `git rebase -i $(git merge-base master HEAD)` (that `$(...)` command just gets git to tell you the commit hash). That will allow you to rewrite just the commits you've made, but doesn't incorporate changes to master since when you started your branch.

------

Personally, for `assert()`s, I prefer to use them in a case like this, where we know that it's always going to be true. To me, the purpose of an `assert()` is to verify an assumption and communicate an assumption. If it's O(1) cost to do so, I prefer to do the assert. However, since in this case it requires adding more code (especially since the diff is small to begin with), I think it's okay not to `assert()`.

VladM1076: Thank you for the information. I will squash those commits for future PRs before I submit them. For this PR I will use the last commit message from my branch as the final commit message. (Fix SequentialListScheduler::IsSequentialInstruction undefined behavi…)

Quincunx271: Okay. FYI, you don't need to get it right _before_ you submit the PR. If you rebase it, you'll just have to force push the branch and the PR will be updated.

paul-mchugh (approved):
This feels like a hacky solution, but I do not object to it since this it resolves an invalid array access bug.  Some time in the future, I would like to write a more robust sequential list scheduler (that is able to tolerate DDG mutations).

Provided that this pr is squashed and merged, I approve.

vangthao95 (approved):
LGTM.

PR #123, opened 2020-09-20
Fixed runspec-wrapper bug where timedout block improvements were not counted

paul-mchugh:
NOTE: ANY B&B COMPILE TIME COSTS THAT WERE GENERATED WITH THE JSON RUNSPEC WRAPPER ARE TAINTED AND SHOULD NOT BE USED


vangthao95 (approved):
LGTM.

PR #124, opened 2020-09-22
Aco improvements, ACO event based logging, and other runspec improvements

paul-mchugh:


paul-mchugh:
Do: reformat

paul-mchugh:
Do: Reformat

Quincunx271:
The checks won't re-run after the `Do: Reformat`. This is a limitation of GitHub's workflows. The easiest way to get it to re-check would be to add a dummy commit, push, then revert the dummy commit and force push.

lib/Scheduler/aco.cpp
```diff
@@ -115,6 +115,25 @@ double ACOScheduler::Score(SchedInstruction *from, Choice choice) {
...
+  if (!OldSched)
+    return true;
+  else if (!NewSched)
+    return false;
+  else if (!rgn_->IsSecondPass())
+    return NewSched->GetCost() < OldSched->GetCost();
+  else
+    return (NewSched->GetSpillCost() <= OldSched->GetSpillCost()) &&
+           (NewSched->GetCost() <  OldSched->GetCost());
+//           (NewSched->GetCrntLngth() <  OldSched->GetCrntLngth());
```
Quincunx271: Should this commented out code be removed? Why is it here?

lib/Scheduler/aco.cpp
```diff
@@ -398,6 +424,11 @@ FUNC_RESULT ACOScheduler::FindSchedule(InstSchedule *schedule_out,
     writePheromoneGraph("iteration" + std::to_string(iterations));
     iterations++;
   }
+
+  Logger::Event(IsPostBB ? "ACOpost_sched_complete" : "ACO_sched_complete",
+                "cost", bestSchedule->GetCost(),
+                "iterations", iterations,
+                "improvement", InitialCost-bestSchedule->GetCost());
```
Quincunx271: The other `Logger::Event(...)` calls would format the event_id as `AcoSchedComplete` or `ACOSchedComplete`.

Another thing to note: we could do this conditional event_id. On the other hand, we could just do `ACOSchedComplete` with another parameter like `"pass", IsPostBB ? "normal" : "post_bb"`. I don't know which is the better strategy.

include/opt-sched/Scheduler/aco.h
```diff
@@ -33,7 +33,7 @@ class ACOScheduler : public ConstrainedScheduler {
 public:
   ACOScheduler(DataDepGraph *dataDepGraph, MachineModel *machineModel,
                InstCount upperBound, SchedPriorities priorities,
-               bool vrfySched);
+               bool vrfySched, bool IsPostBB);
```
Quincunx271: Consider using an enum class for `IsPostBB` instead of a `bool`. When you see a function call, `ACOScheduler(a, b, c, d, e, true, false)` is harder to understand than `ACOScheduler(a, b, c, d, e, true, ACOScheduler::PreBB)`.

lib/Scheduler/aco.cpp
```diff
@@ -536,9 +567,9 @@ void ACOScheduler::writePheromoneGraph(std::string Stage) {
       OutPath + "/" + dataDepGraph_->GetDagID() + "@" + Stage + ".dot";
   FILE *Out = fopen(FullOutPath.c_str(), "w");
   if (!Out) {
-    Logger::Info("Could now open file to write pheromone display at %s."
-                 " Skipping.",
-                 FullOutPath.c_str());
+    Logger::Error("Could now open file to write pheromone display at %s."
+                  " Skipping.",
+                  FullOutPath.c_str());
```
Quincunx271, line 572: I think we wanted to use `llvm::report_fatal_error` for this. It would be something like `llvm::report_fatal_error(llvm::Twine("Could ... Skipping. File: ").concat(FullOutPath), false)` or `llvm::report_fatal_error("Could ... Skipping. File: " + FullOutPath, false)`.

Also, there was a typo: "Could **now**" --> "Could _not_"

Quincunx271 (changes requested):

lib/Scheduler/aco.cpp
paul-mchugh: I used Logger::Error, because this is not a fatal error.  This should not crash the program.  I just wanted to inform the user that an error occurred.  Since this function only provides debugging information regarding the pheromone table the scheduler is able to continue functioning, albeit without extra debugging info.

lib/Scheduler/aco.cpp
paul-mchugh: I can change the event ID format, but adding a pass parameter would be a bad idea since there can be both a preBB and postBB run of the ACO Scheduler

include/opt-sched/Scheduler/aco.h
paul-mchugh: I don't anticipate adding more intra-pass invocations of the ACO Scheduler.  runACO is a short private method that is only called in two places.  It is easy to determine which what the boolean IsPostBB argument does by looking at the function signature.  I don't think that adding a enum that will only be used once or twice is beneficial.

lib/Scheduler/aco.cpp
Quincunx271: Thanks, I goofed when reading the code.

Quincunx271 (approved):
LGTM

vangthao95 (approved):

PR #125, opened 2020-09-22
Add a JSON -> info log conversion script

Quincunx271:
Translates logs with the EVENT: + JSON format to the old INFO: format.

This is intended to be used for script verification purposes. An updated
script's results can be compared with the un-updated script by
translating the JSON logs and passing it through the older script.

----------

I still need to do more testing of this script.


Quincunx271:
I've tested, and there are no apparent functional changes.

vangthao95:
I also tested and this works great. LGTM.

Quincunx271:
TODO: translation for Paul's latest merge, and add a better error message for unknown event_ids

PR #126, opened 2020-09-26
Fixed bug in enumerator algorithm

jrbyrnes:
The intended logic of the two-pass algorithm is to find a good / optimal value of RP cost in the first pass, then iterate through minimum possible schedule lengths until we find a schedule that matches that predetermined RP cost.

Before this fix, the second pass would find the first pass' RP cost value for some schedule length, then continue to explore the rest of the tree for the given schedule length. Only after exploring to completion, or timing out, would the alg determine it had met the second pass' conditions. This additional exploring is wasted time as the conditions for finding good RP costs are worse in second pass compared to first pass, and I have seen no evidence that additional exploring in the second pass leads to better results. Further, in some cases, the additional exploring represents the overwhelming majority of second pass exploration (in the pic, the second pass finds desired schedule in 2ms, but continues exploring for another 235ms).

Screenshot from 2020-09-24 15-19-08

This PR includes the code to stop exploring at the exact moment we find a min length schedule at the first pass' RP cost value.

### DIFF
--- a/lib/Scheduler/bb_spill.cpp
+++ b/lib/Scheduler/bb_spill.cpp
@@ -864,7 +864,7 @@ InstCount BBWithSpill::UpdtOptmlSched(InstSchedule *crntSched,
 
   if (crntCost < GetBestCost()) {
 
-    if (crntSched->GetCrntLngth() > schedLwrBound_)
+    if (crntSched->GetCrntLngth() > schedLwrBound_ && !(isTwoPassEnabled()))
       Logger::Info("$$$ GOOD_HIT: Better spill cost for a longer schedule");
 
     SetBestCost(crntCost);
@@ -873,7 +873,7 @@ InstCount BBWithSpill::UpdtOptmlSched(InstSchedule *crntSched,
     enumBestSched_->Copy(crntSched);
     bestSched_ = enumBestSched_;
   }
-
+  
   return GetBestCost();
 }
 /*****************************************************************************/
--- a/lib/Scheduler/enumerator.cpp
+++ b/lib/Scheduler/enumerator.cpp
@@ -826,7 +826,7 @@ void AppendAndCheckSuffixSchedules(
   Logger::Info("Found a concatenated schedule with node instruction %d",
                crntNode_->GetInstNum());
 #endif
-  if (newCost < oldCost) {
+  if (newCost < oldCost && !(rgn_->IsSecondPass())) {
 #if defined(IS_DEBUG_SUFFIX_SCHED)
     Logger::Info("Suffix Scheduling: Concatenated schedule has better "
                  "cost %d than best schedule %d!",
@@ -2040,13 +2040,25 @@ bool LengthCostEnumerator::WasObjctvMet_() {
   InstCount crntCost = GetBestCost_();
 
   InstCount newCost = rgn_->UpdtOptmlSched(crntSched_, this);
-  assert(newCost <= GetBestCost_());
-
-  if (newCost < crntCost) {
-    imprvmntCnt_++;
+  if (rgn_->IsSecondPass()) {
+    // for second pass -- newCost is value of RP for sched
+    if (newCost == rgn_->getSpillCostConstraint() && 
+        trgtSchedLngth_ < getBestSchedLength_()) {
+      imprvmntCnt_++;
+    }
+    
+    return newCost == rgn_->getSpillCostConstraint();
   }
 
-  return newCost == costLwrBound_;
+  else {
+    assert(newCost <= GetBestCost_());
+    // for first pass & non-twoPass alg -- newCost is weighted sum cost for sched
+    if (newCost < crntCost) {
+      imprvmntCnt_++;
+    }
+
+    return newCost == costLwrBound_;
+  }
 }
 /*****************************************************************************/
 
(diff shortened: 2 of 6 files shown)


jrbyrnes:
Do: Reformat

jrbyrnes:
I realize that my previous changelogs were not clean -- refer to this commit to get a comprehensive view of changes.

lib/Scheduler/bb_spill.cpp
```diff
@@ -852,6 +852,8 @@ InstCount BBWithSpill::UpdtOptmlSched(InstSchedule *crntSched,
   InstCount crntCost;
   InstCount crntExecCost;
 
+  //std::pair<InstCount, InstCount> retPair;
```
vangthao95, line 855: This should be removed if it isn't needed.

lib/Scheduler/bb_spill.cpp
```diff
@@ -873,6 +876,11 @@ InstCount BBWithSpill::UpdtOptmlSched(InstSchedule *crntSched,
     enumBestSched_->Copy(crntSched);
     bestSched_ = enumBestSched_;
   }
+  
+  if (IsSecondPass())
+  {
+    return optmlSpillCost_;
+  }
```
vangthao95, line 883: Remove braces on simple if statements. https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements

vangthao95 (changes requested):

lib/Scheduler/sched_region.cpp
```diff
@@ -239,6 +239,9 @@ FUNC_RESULT SchedRegion::FindOptimalSchedule(
     CmputNormCost_(lstSched, CCM_DYNMC, hurstcExecCost, true);
     hurstcCost_ = lstSched->GetCost();
 
+    // Get unweighted spill cost for Heurstic list scheduler
+    HurstcSpillCost_ = lstSched->GetSpillCost();
+
```
paul-mchugh, line 244: getSpillCost() has issues when used with SLIL.  It gives you the SLIL cost without subtracting the SLIL bound.  Currently as a consequence of a bug, The lower bound on SLIL is only subtracted from the value returned by GetCost, and is not included in the value returned by GetSpillCost.  (This is not an issue for other cost functions because no other cost functions have no RP cost lower bound).

paul-mchugh (changes requested):
I am not sure that the criteria you put for MetObjective are correct.  Also if you are going to rely on the GetSpillCost function you should know that It does not include have its lower bound subtracted out.  I also have some minor naming quibbles regarding the function "getFirstPassSpillCost()". Please view the in code comments for further details.

lib/Scheduler/enumerator.cpp
```diff
@@ -2040,13 +2040,25 @@ bool LengthCostEnumerator::WasObjctvMet_() {
...
+  if (rgn_->IsSecondPass()) {
+    // for second pass -- newCost is value of RP for sched
+    if (newCost <= rgn_->getFirstPassSpillCost() && 
+        trgtSchedLngth_ <= getBestSchedLength_()) {
+      imprvmntCnt_++;
+    }
 
-  if (newCost < crntCost) {
-    imprvmntCnt_++;
+    return newCost == rgn_->getFirstPassSpillCost();
```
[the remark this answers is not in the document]
jrbyrnes: One of the preconditions for this code is that the ILP constraint is met -- in other words, the schedule that we have at this given targetLength is a valid schedule. Thus, the only condition we need to check for second pass is if the RP value is the value we found in the first pass. If you look at my changes to updtOptmlSchedule, you'll see that, in the second pass, this function returns the RP value. Thus, we are comparing the second pass RP to first pass RP both to determine optimality and improvement. I'm not thrilled with how updtOptmlSched is coded in this PR, but I couldn't think of another way without introducing updtOptmlSchedScndPss to do second pass specific stuff -- which starts to blur the line between this commit (where weighted cost function is still king) and my next PR wherein everything is split up.

You're right that the condition to increment imprvmntCnt_ is wrong. Essentially, I think it should only happen if (trgtLength < getBestSchedLength() && newCost == getFirstPassSpillCost()) -- that is, if we found a same-RP schedule that is shorter than the one passed to the second pass. I see the confusion about the RP condition here -- in this PR, I decided less than or equal (rather than equal) on the very rare - if not impossible - chance that we find a better RP in the second pass. However, the entire premise of this PR is that we will not find a better RP in the second pass than we found in the first. Either way, though, we shouldn't be updating the firstPassRP - and, on the first encounter of a schedule wherein the second pass RP == firstPassRP, we stop traversing (another premise of this PR) - so I don't think we'll ever be concerned with updating the firstPassSpillCost.

lib/Scheduler/sched_region.cpp
jrbyrnes: I saw something like this in my other branch, and fixed it there. I didn't see it come up in testing for this PR though -- I'll double check and apply fix if needed.

jrbyrnes: We have actually elected to use weighted cost function for SLIL in two-pass for this PR, this is reflected in my most recent push.

PR #128, opened 2020-10-06
Add workflow retriggering

Quincunx271:
This adds a way to rerun the checks when they don't want to run (e.g. after a `Do: Reformat`) via the comment code `Do: Run Checks`.


Quincunx271:
Seems that the C++ build didn't run for some reason. I need to investigate that

Quincunx271:
Do: Run Checks

Quincunx271:
Hmm. I suspect that didn't work because the updated workflow files are in this PR and not on the master branch. The github bot is supposed to respond with a comment like https://github.com/Quincunx271/Sandbox/pull/1#issuecomment-703974139 for each check.

Quincunx271:
Another problem: this checks the _master_ branch, not the PR. I need to do more work here

Quincunx271:
An idea: perhaps this check rerunning would be best as its own workflow file rather than duplicating the code across each check.

Quincunx271:
I can't check whether the comment response works without merging this first, so I think this PR is ready to merge (commit squashing will happen, of course).

PR #129, opened 2020-10-13
Removal of weighted cost function for two-pass

jrbyrnes:
In an effort to align the logic of our algorithm more closely with the algorithm defined in our papers, this PR implements three main things:

1. Instead of using a cost function that is weighted between schedLength and RP, we make our decisions based on either the "raw" value of schedLength or RP or both (both meaning multiple conditional statements rather than a value that combines both sources into one).
2. We stop exploring the tree in the second pass after we have found a schedule that corresponds with the first pass' value of spillCostFUunc
3. If, after introducing latency information to the instructions, the max latency is 1, we skip the second pass as it will just be an extension of the first pass.

The functions where we make use of weighted cost function are, UpdtOptmlSchedule, ChkCostFsblty, WasObjctvMet, and ChkCostDomntn. These have been split up into functions for frstPss, scndPss, and wghtd to implement pass specific code. This was done to enhance code clarity.

Finally, appendAndCheckSuffixes makes use of the weighted cost function, however, it seems this feature is no longer used / needs to be maintained. As a result, this PR does not attempt to remove the weighted cost from this function, and, in fact, this code ensures appendAndCheckSuffixes does not work with the two-pass algorithm. That said, I have continued to maintain the weightedCost value in our two-pass algorithm, so reactivating appendAndCheckSuffixes for the two-pass algorithm will be a simple manner of using the pass-dependent methods to get the weightedCost and invoking the unchanged logic of appendAndCheckSuffixes which makes use of the weighted cost.

### DIFF
--- a/example/optsched-cfg/sched.ini
+++ b/example/optsched-cfg/sched.ini
@@ -189,6 +189,7 @@ REGIONS_TO_SCHEDULE fft1D_512:114
 
 # Whether to use suffix concatenation. Disabled automatically if
 # history domination is disabled.
+# WARNING: OUT OF DATE - GAURANTEED TO FAIL ON TWOPASS VERSION OF ALGORITHM
 ENABLE_SUFFIX_CONCATENATION NO
 
 # Whether to apply the node superiority graph transformation.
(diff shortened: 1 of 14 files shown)


jrbyrnes:
Do: Reformat

PR #130, opened 2020-10-14
Fix missing LLVM GPG key

Quincunx271:
The clang-tidy check uses apt-get to install LLVM components. This adds the missing GPG key.


Quincunx271:
Hmm. I thought that was the problem. But apparently there's something still wrong about this

PR #131, opened 2020-10-19
Add ArrayRef2D<T> for a 2D view of a single array

Quincunx271:
When working with a 2D array/matrix from a single allocation,
ArrayRef2D<T> and MutableArrayRef2D<T> provides a simpler way to access
a 2D location in this matrix.

----

Added Paul as a reviewer because I saw some 2d table accessing in the ACO implementation.


paul-mchugh (approved):
LGTM, Currently I am working on an old branch of the scheduler, so I do not intend to update ACO to use the 2d array immediately.  I agree with merging this data structure now.  I will hopefully have the occasion to update ACO  soon.

PR #134, opened 2020-10-28
Split up weighted cost, changes to 2-pass

jrbyrnes:
This PR includes a number of changes to the two-pass enumerator algorithm, namely:

- Split the weighted sum cost into its individual parts (spill cost and schedule length)
- In the second pass, stop exploration if we encounter a schedule with the RP found in the first pass
- Mechanism to ensure we skip the second pass if the max latency <= 1 (and the second pass would just extend the first pass)
- Propagate up spill costs to parent nodes in cases wherein children fail cost probes (this allows better use of history domination based cost pruning)


include/opt-sched/Scheduler/array_ref2d.h
```diff
@@ -102,4 +102,4 @@ template <typename T> class MutableArrayRef2D : public ArrayRef2D<T> {
 } // namespace opt_sched
 } // namespace llvm
 
-#endif
+#endif
```
Quincunx271, line 105: Is your editor removing newlines from the end of files? It's preferred that there is a single newline at the end of each file, so I'd ask that the newlines be added back.

include/opt-sched/Scheduler/graph.h
```diff
@@ -621,4 +621,4 @@ inline DIRECTION DirAcycGraph::ReverseDirection(DIRECTION dir) {
 } // namespace opt_sched
 } // namespace llvm
 
-#endif
+#endif
```
Quincunx271, line 624: Ditto

include/opt-sched/Scheduler/graph_trans.h
```diff
@@ -112,4 +112,4 @@ class StaticNodeSupTrans : public GraphTrans {
 } // namespace opt_sched
 } // namespace llvm
 
-#endif
+#endif
```
Quincunx271, line 115: Ditto

include/opt-sched/Scheduler/lnkd_lst.h
```diff
@@ -798,4 +798,4 @@ void PriorityList<T, K>::InsrtEntry_(KeyedEntry<T, K> *entry,
 } // namespace opt_sched
 } // namespace llvm
 
-#endif
+#endif
```
Quincunx271, line 801: Ditto

lib/Scheduler/bb_spill.cpp
```diff
@@ -779,6 +786,7 @@ FUNC_RESULT BBWithSpill::Enumerate_(Milliseconds startTime,
   FUNC_RESULT rslt = RES_SUCCESS;
   int iterCnt = 0;
   int costLwrBound = 0;
+  // int SpillCostLwrBound = cmputSpillCostLwrBound();
```
Quincunx271, line 789: Commented out code?

lib/Scheduler/bb_spill.cpp
```diff
@@ -847,8 +855,101 @@ FUNC_RESULT BBWithSpill::Enumerate_(Milliseconds startTime,
...
+}
+
+/*****************************************************************************/
+
+SmallVector<InstCount, 4>
+BBWithSpill::UpdtOptmlSchedFrstPss(InstSchedule *crntSched,
+                                   LengthCostEnumerator *) {
+  InstCount crntCost;
+  InstCount crntExecCost;
+  InstCount TmpSpillCost;
```
Quincunx271, line 880: Move these to the definition. Having variable declarations at the top of a function is an old C-ism (for old versions of C) which makes code harder to read.

lib/Scheduler/bb_spill.cpp
```diff
@@ -847,8 +855,101 @@ FUNC_RESULT BBWithSpill::Enumerate_(Milliseconds startTime,
...
+  llvm::SmallVector<InstCount, 4> retVec;
+
+  //  crntCost = CmputNormCost_(crntSched, CCM_DYNMC, crntExecCost, false);
```
Quincunx271, line 883: Commented out code? There's other commented out code in this function too.

lib/Scheduler/bb_spill.cpp
```diff
@@ -847,8 +855,101 @@ FUNC_RESULT BBWithSpill::Enumerate_(Milliseconds startTime,
...
+  return retVec;
+}
+
+/*****************************************************************************/
+
+SmallVector<InstCount, 4>
+BBWithSpill::UpdtOptmlSchedWghtd(InstSchedule *crntSched,
+                                 LengthCostEnumerator *) {
+  llvm::SmallVector<InstCount, 4> retVec;
+
```
Quincunx271, line 952: There's a lot of duplicated code between these three functions. Maybe pull out the duplicated code pieces into common functions?

lib/Scheduler/bb_spill.cpp
```diff
@@ -899,32 +1002,131 @@ void BBWithSpill::SetupForSchdulng_() {
...
+    return true;
+  }
+
+  RPCost = TmpSpillCost;
+  return false;
+}
+
+/*****************************************************************************/
+
+bool BBWithSpill::ChkCostFsbltyWghtd(InstCount trgtLngth, EnumTreeNode *node) {
```
Quincunx271, line 1099: Similarly, there's duplicated code between these 3 functions. There has to be some way to reduce the duplication without cluttering it with a bunch of conditionals.

lib/Scheduler/hist_table.cpp
```diff
@@ -467,33 +493,106 @@ static bool doesHistoryPeakCostDominate(InstCount OtherPrefixCost,
...
+      return true;
+
+    // if the above condition fails (e.g. HistPrefixSpillCost >=
+    // HistSuffixRPCost) then we will arrive at this assert. We also know that
+    // HistSuffixRPCost >= TrgtSpillCost due to outer condition. Thus, this
+    // would imply that we are currently in a condition where
+    // HistPrefixSpillCost > TrgtSpillCost. Based on the way we store history
+    // schedules, it is impossible to reach this condition. This assert should
+    // never be reached
+    assert(false);
```
Quincunx271, line 526: llvm::report_fatal_error? At a minimum, this should be `assert(false && "Some message making it possible to figure out which assert fired and giving a bit of information)`

lib/Scheduler/list_sched.cpp
```diff
@@ -156,4 +156,4 @@ void ListScheduler::UpdtRdyLst_(InstCount cycleNum, int slotNum) {
     rdyLst_->AddList(lst2);
     lst2->Reset();
   }
-}
+}
```
Quincunx271, line 159: Shouldn't remove the trailing newline

lib/Scheduler/sched_region.cpp
```diff
@@ -414,13 +427,16 @@ FUNC_RESULT SchedRegion::FindOptimalSchedule(
...
       bestCost_ = hurstcCost_;
+      BestSpillCost_ = HurstcSpillCost_;
     }
     // B) Heuristic was never run. In that case, just use ACO and run with its
     // results, into B&B.
     else if (!HeuristicSchedulerEnabled) {
       bestSched = bestSched_ = AcoSchedule;
       bestSchedLngth_ = AcoScheduleLength_;
       bestCost_ = AcoScheduleCost_;
+      // need to set spillCost
```
Quincunx271, line 438: Is this a TODO comment? Was this addressed?

lib/Scheduler/sched_region.cpp
```diff
@@ -429,6 +445,7 @@ FUNC_RESULT SchedRegion::FindOptimalSchedule(
       bestSched = bestSched_;
       bestSchedLngth_ = bestSched_->GetCrntLngth();
       bestCost_ = bestSched_->GetCost();
+      // need to set spillCost
```
Quincunx271, line 448: Ditto

lib/Scheduler/sched_region.cpp
```diff
@@ -832,6 +854,7 @@ void SchedRegion::CmputLwrBounds_(bool useFileBounds) {
     UseFileBounds_();
 
   costLwrBound_ = CmputCostLwrBound();
+  // SpillCostLwrBound_ = getSpillCostLwrBound();
```
Quincunx271, line 857: Commented out code

include/opt-sched/Scheduler/sched_region.h
```diff
@@ -77,11 +86,28 @@ class SchedRegion {
...
+  UpdtOptmlSchedFrstPss(InstSchedule *crntSched,
+                        LengthCostEnumerator *enumrtr) = 0;
+
+  virtual llvm::SmallVector<InstCount, 4>
+  UpdtOptmlSchedScndPss(InstSchedule *crntSched,
+                        LengthCostEnumerator *enumrtr) = 0;
+
+  virtual llvm::SmallVector<InstCount, 4>
+  UpdtOptmlSchedWghtd(InstSchedule *crntSched,
+                      LengthCostEnumerator *enumrtr) = 0;
```
Quincunx271, line 106: If there's a common reason why all of these functions' results have a small size of `4`, I'd prefer that the `4` be pulled out into a constant. If the reason can be explained with a comment on this constant, all the better, but that's not necessarily needed if the number is arbitrary.

Quincunx271 (changes requested):
I recognize that you may not be finished with the PR, but I noticed some things that I wanted to comment on. The low-level comments are inline, but I have a high-level comment as well:

In these changes, you're splitting the cost functions into 3:
* First-pass RP-only
* Second-pass ILP-only but maintaining occupancy
* Single-pass weighted sum

To do so, you're having the original cost function branch to determine which of these three cost functions to evaluate. Would it perhaps be better to instead have the cost function be a `std::function<>` member or new abstract `CostFunction` member? Going the `CostFunction` route would look something like:

```
// I currently don't have the strongest model on what needs to be in here,
// but pretend something like this suffices
class CostFunction {
public:
  virtual ~CostFunction() = default;
  virtual Cost computeCost() const = 0;
  virtual bool checkCost(...) const = 0;
};
```

And there would be 3 derived classes for the 3 variations.

In fact, I wonder if there should actually be separate derived classes for each of our cost functions (PRP, PERP, SLIL), as well as a combiner to get a WeightedCostFunction and a SecondPassCostFunction, but that could easily lead to significant code duplication.

Despite this strategy requiring a virtual call, it's possible that it might not be slower than the current branchy approach because the branch predictor can predict virtual calls and because it might have better icache performance.

lib/Scheduler/bb_spill.cpp
Quincunx271: Although, if this is just existing code, it's okay to leave as-is. I'm not sure because I think this is copied code. It's safe to move them to the definition, though, and I prefer to do that when I touch the existing code.

lib/Scheduler/hist_table.cpp
jrbyrnes: I figured someone would say something about this one -- I'll go with the fatal_error -- thanks

include/opt-sched/Scheduler/sched_region.h
vangthao95: 4 was just an arbitrary number that was big enough for returning what we need such as register pressure, schedule length, and possibly clustering in my project.

lib/Scheduler/sched_region.cpp
jrbyrnes: Yes this was meant to be a TODO -- good catch on this. Thank you

lib/Scheduler/sched_region.cpp
```diff
@@ -320,6 +325,9 @@ FUNC_RESULT SchedRegion::FindOptimalSchedule(
   else
     CmputLwrBounds_(false);
 
+  // Set the lowerbound of spill cost (need to CmptCostLWrBound() first)
+  SpillCostLwrBound_ = getSpillCostLwrBound();
```
vangthao95, line 329: This might be redundant. It is setting `SpillCostLwrBound_ = SpillCostLwrBound_`

jrbyrnes: Yeah I think this is redundant. The reason why, however, is that CmputLwrBounds_ is coded weirdly. It has the side effect of setting the spill LB yet it returns the cost LB. Would it be better if we set both, return neither, and rename the function cmputeAndSetLB. This would help avoid the redundancy error in the future.

vangthao95: I am fine with renaming it `cmputeAndSetLB` since that would make more sense.

lib/Scheduler/sched_region.cpp
```diff
@@ -386,8 +438,12 @@ FUNC_RESULT SchedRegion::FindOptimalSchedule(
     if (AcoTime > 0)
       Logger::Info("ACO_Time %d", AcoTime);
 
+    InstCount acoExecCost;
+
     AcoScheduleLength_ = AcoSchedule->GetCrntLngth();
+    CmputNormCost_(AcoSchedule, CCM_DYNMC, acoExecCost, true);
```
vangthao95, line 444: This is redundant. ACO already called this so AcoSchedule should already have all costs related information.

lib/Scheduler/enumerator.cpp
```diff
@@ -2017,6 +2046,17 @@ FUNC_RESULT LengthCostEnumerator::FindFeasibleSchedule(InstSchedule *sched,
...
   rgn_ = rgn;
   costLwrBound_ = costLwrBound;
+  SpillCostLwrBound_ = rgn_->getSpillCostLwrBound();
+
+  this->setIsSecondPass(rgn_->IsSecondPass());
+  this->setIsTwoPass(rgn_->isTwoPassEnabled());
+
+  if (rgn_->IsSecondPass())
+    TrgtSpillConstraint_ = rgn_->getSpillCostConstraint();
+
```
vangthao95, line 2056: Whitespace

lib/Scheduler/enumerator.cpp
```diff
@@ -2031,22 +2071,76 @@ FUNC_RESULT LengthCostEnumerator::FindFeasibleSchedule(InstSchedule *sched,
...
   assert(GetBestCost_() >= 0);
 
   if (WasSolnFound_() == false) {
     return false;
   }
 
+  if (!rgn_->isTwoPassEnabled())
+    return WasObjctvMetWghtd_();
+
```
vangthao95, line 2085: Whitespace

lib/Scheduler/bb_spill.cpp
```diff
@@ -847,21 +854,62 @@ FUNC_RESULT BBWithSpill::Enumerate_(Milliseconds startTime,
...
 
-  //#ifdef IS_DEBUG_SOLN_DETAILS_2
-  Logger::Info(
-      "Found a feasible sched. of length %d, spill cost %d and tot cost %d",
-      crntSched->GetCrntLngth(), crntSched->GetSpillCost(), crntCost);
-  //  crntSched->Print(Logger::GetLogStream(), "New Feasible Schedule");
-  //#endif
+  Logger::Event("feasible_sched_found", "length", crntSched->GetCrntLngth(),
+                "spill_cost", crntSched->GetSpillCost(), "cost", crntCost);
+  crntSched->Print(Logger::GetLogStream(), "New Feasible Schedule");
```
vangthao95, line 865: I think the above feasible sched found event log is good enough. We don't need to print the schedule also.

vangthao95 (changes requested):
Just some whitespace left and one small change that I want to request then it looks good.

lib/Scheduler/sched_region.cpp
```diff
@@ -503,7 +527,12 @@ FUNC_RESULT SchedRegion::FindOptimalSchedule(
     Milliseconds enumStart = Utilities::GetProcessorTime();
     if (!isLstOptml) {
       dataDepGraph_->SetHard(true);
-      rslt = Optimize_(enumStart, rgnTimeout, lngthTimeout);
+      if (IsSecondPass() && dataDepGraph_->GetMaxLtncy() <= 1)
+        Logger::Info("Problem size not increased after introducing latencies, "
+                     "skipping enumeration");
```
paul-mchugh, line 532: You might want to clarify that wea re skipping enumeration in the second pass

paul-mchugh (commented):
I was asked to review the changes to sched_region.cpp.  I have observed no issues on which i would object to merging this PR.

My one comment is that the message "Problem size not increased after introducing latencies, skipping enumeration" is some what unclear.  I think that the message should be clarified to state that it is only skipping enumeration in the second pass.

PR #135, opened 2020-10-29
Refactoring out the allocator mechanism in LinkedList

Quincunx271:
This fixes LinkedList node removal / edge removal. Previously, I merged a PR which allowed nodes to be removed from a LinkedList, but I did not update PriorityList. PriorityList has its own allocator mechanism, which meant that removing nodes from the PriorityList would fail and cause undefined behavior.

This change unifies the allocation mechanisms of LinkedList and PriorityList by pulling it out into an `EntryAllocator` interface.

This should not have a negative runtime impact, because we previously had a virtual call to allocate an entry, and we still have a virtual call, but we also have fewer branches while allocating and the sizes of LinkedList and PriorityList are smaller.

----

I plan to squash these two commits together. I kept them separate because that might possibly be easier to review.


paul-mchugh (approved):
LGTM

PR #140, opened 2021-01-04
Abstracted log parser: uniformity in analysis scripts, more straightforward analysis

Quincunx271:
**Note:** The code has changed significantly since this comment with a lot of simplification. I'm keeping this comment here because it's the history of the development of this branch, but the latest comment at the end of the PR is more reflective of how the code looks.

----

<details>
<summary>The old text</summary>

This adds an abstraction on the JSON logs which makes it easier to write scripts to analyze runs.

I've realized that I find analyzing runs to be stressful because the output directory structure of different benchmark suites is different and because the actual writing of analysis code was difficult. This addresses these issues by creating benchmark suite "importers" and by simultaneously being a library for analyzing our logs.

A quick preview of using this can be seen in util/gt_analysis/gt_cost.py. An excerpt (most of the file, TBH):

```python3
class RpGtCost(analyze.Analyzer):
    '''
    Computes the compile time cost of Register Pressure Graph Transformations.
    '''

    POSITIONAL = {
        'nogt': 'Logs without Graph Transformations',
        'gt': 'Logs with Graph Transformations'
    }.items()

    def run_bench(self, args):
        nogt = args[0]
        gt = args[1]

        gt_cost = sum(_rp_gt_elapsed(blk) for blk in gt.all_blocks())
        gt_cost_before_enum = sum(_cost_before_enum(blk) for blk in gt.all_blocks())
        nogt_cost_before_enum = sum(_cost_before_enum(blk) for blk in nogt.all_blocks())
        gt_total_cost = gt_cost_before_enum - nogt_cost_before_enum

        self.stat(
            gt_cost=gt_cost,
            gt_total_cost=gt_total_cost,
        )


if __name__ == '__main__':
    analyze.main(RpGtCost)
```

Writing the analysis code is much easier because you mostly only have to write the one `run_bench()` function (which runs the analysis over each benchmark as well as once over everything) or a `run()` function (which would instead run the analysis only once over everything).

Note that the main function is implemented by the library, which would tend to make all analysis scripts using this library have a uniform calling syntax and uniform output syntax (CSV for the time being).

----

Unfortunately, for the larger benchmarks such as FP2006, the log files can take a while to parse (close to 30 seconds per file) and use a lot of memory (two files together took close to 10 GB), which appears mostly unavoidable if we want to use Python. To help mitigate this, existing analysis scripts can be combined together in a "script file", which will only parse logs once at the start instead of for every analysis:

```python3
#!/usr/bin/env python3

# import all the analysis classes
from analyze.script import *

run(RpGtCost, _0, _1)
# Or, to only run over blocks matching some pattern, e.g.:
# Only blocks which were enumerated
# where we also found something better than the heuristic
run(SomeOtherAnalysis, _1) \
    .keep_if(lambda blk: 'Enumerating' in blk) \
    .keep_if(lambda blk: blk.single('BestResult')['cost'] < blk.single('HeuristicResult')['cost'])
```

</details>


Quincunx271:
The "scripting" mechanism is fundamentally broken. It should be part of the main ecosystem. That is to say, I should be able to do the `keep_if` filtering directly on the benchmarks, possibly on the commandline, and running multiple analyses together should be easy just by writing an analysis class.

Quincunx271:
This current state is dramatically simpler. It's now just a library for easing writing these scripts. Features:

 - Types which extract the logfiles (`Logs`, `Benchmark`, `Block`)
 - Utilities for writing main:
   - `analyze.parse_args(parser)` instead of `parser.parse_args()` gives a few arguments for free, and the ability to automatically load logs specified via `analyze.parse_logs` (see next).
   - `parser.add_argument(... type=analyze.parse_logs)` will parse the argument into the `Logs` data type, provided that `analyze.parse_args(...)` is used instead of `parser.parse_args()`. (see next for why)
   - `--benchsuite` is automatically added, providing a way to specify that the logs are PlaidML, SHOC, or SPEC. (this is why `analyze.parse_args(...)` exists)
   - `--keep-blocks-if` provides a very basic JSON format filter to only consider blocks following a certain desired predicate
 - Utility for running the analysis over each benchmark, gathering the results into a dict

Quincunx271:
Okay. This is v3 of it, and it's overall a lot better. See util/analyze/lib/compile_times.py for a good example of what using this is like: https://github.com/Quincunx271/OptSched/blob/abstracted-log-parser/util/analyze/lib/compile_times.py

```python
#!/usr/bin/env python3

import csv
import re
import argparse
import sys

import analyze
from analyze import Block, foreach_bench


def _block_time(block: Block):
    end = block.single('ScheduleVerifiedSuccessfully')['time']
    start = block.single('ProcessDag')['time']
    return end - start


def instruction_scheduling_time(logs):
    return sum(_block_time(blk) for blk in logs)


def total_compile_time_seconds(logs):
    last_logs = logs.benchmarks[-1].blocks[-1].raw_log
    m = re.search(r'(\d+) total seconds elapsed', last_logs)
    assert m, \
        'Logs must contain "total seconds elapsed" output by the SPEC benchmark suite'

    return m.group(1)


if __name__ == '__main__':
    parser = argparse.ArgumentParser()
    parser.add_argument('--variant', choices=('sched', 'total'),
                        help='Which timing variant to use')
    parser.add_argument('logs', help='The logs to analyze')
    args = analyze.parse_args(parser, 'logs')

    fn = total_compile_time_seconds if args.variant == 'total' else instruction_scheduling_time
    results = foreach_bench(fn, args.logs)
    writer = csv.DictWriter(sys.stdout, fieldnames=results.keys())
    writer.writeheader()
    writer.writerow(results)
```

The gist of it is that writing analysis can be split into two parts: the analysis functions and the runner. In this case, `instruction_scheduling_time` and `total_compile_time_seconds` are the analysis functions, and block in `if __name__ == '__main__':` is the runner.

Analysis functions use the types defined in util/analyze/_types.py, which provide a convenient abstracted view of the log files (abstracted in that it doesn't treat it as a string, but as a sequence of `Block`s that have attributes for the events, etc.).

The runner greatly benefits by using `analyze.parse_args` to add a couple arguments (`--benchsuite` to specify whether the blocks are `spec` (CPU2006), `plaidml`, `shoc`, etc.; `--keep-blocks-if` which allows for a simple JSON filter to only keep blocks matching certain criteria). This function will help parse any logfile arguments on the `parser`, parsing using the correct benchmark format.

-----

I've been using this style of analysis script for a while, and it's greatly improved my ability to write analysis and understand my analysis.

Quincunx271:
FWIW, the performance problems are not an issue with this PR, but an issue with parsing the EVENT logs. I'm not 100% sure why; it's possible that `json.loads` is an issue, or it's possible that all the `dict`s is the issue

util/analyze/_utils.py
```diff
@@ -0,0 +1,20 @@
...
+        The keys are the benchmark names.
+        The run for the entire thing has a key of 'Total'
+    '''
+
+    benchmarks = zip(*[log.benchmarks for log in logs])
+
+    return {
+        'Total': analysis_f(*logs),
+        **{bench[0].name: analysis_f(*bench) for bench in benchmarks}
+    }
```
paul-mchugh, line 20: This could be a cause of performance problems, this code calls the analysis function on all the logs in line 18, then subsequently calls it for each benchmark in line 19. This likely doubles runtime(you said that runtimes were long).  Would it be possible to combine the analysis of the benchmarks to form the analysis of the total?

Quincunx271: Most of the performance problem is in loading the log files. Actually running the analysis is fairly fast. However, you're absolutely right that this is a good opportunity for a performance improvement.

------

I just measured performance for a script that just loads the logs and does nothing with it against one that does some useful analysis. Just loading logs: `0m28.228s`. Calculating some stats on the logs after loading it: `0m36.344s`. So most of the time is spent loading the logfile, but it could still be reasonable to optimize this (the calculated stats aren't the most extreme either)

paul-mchugh (approved):
LGTM

util/analyze/_utils.py
Quincunx271: It was easy enough to fix, so I went ahead and did so.

PR #141, opened 2021-01-09
Add workflow retriggering

Quincunx271:
A continuation of #128 where the manual triggering of checks is tested.


Quincunx271:
Do: Run Checks

Quincunx271:
Do: Run Checks

Quincunx271:
Well that's annoying. Every time I want to make small changes, I have to commit to _master_. I'll probably have to do some experimentation and rollback master after my experimentation is done.

PR #142, opened 2021-01-19
Aco costs improvement

paul-mchugh:
This PR updates the ACO code and some code in bb_spill.cpp to better handle situations where ACO is run in 2 pass mode.  Adds a version of ACO that uses two cost functions in one pass.

### DIFF
--- a/include/opt-sched/Scheduler/aco.h
+++ b/include/opt-sched/Scheduler/aco.h
@@ -52,7 +52,8 @@ private:
   pheromone_t &Pheromone(SchedInstruction *from, SchedInstruction *to);
   pheromone_t &Pheromone(InstCount from, InstCount to);
   double Score(SchedInstruction *from, Choice choice);
-  bool shouldReplaceSchedule(InstSchedule *OldSched, InstSchedule *NewSched, bool IsGlobal);
+  bool shouldReplaceSchedule(InstSchedule *OldSched, InstSchedule *NewSched,
+                             bool IsGlobal);
   DCF_OPT ParseDCFOpt(std::string opt);
 
   void PrintPheromone();
--- a/include/opt-sched/Scheduler/aco.h
+++ b/include/opt-sched/Scheduler/aco.h
@@ -54,7 +54,7 @@ private:
   double Score(SchedInstruction *from, Choice choice);
   bool shouldReplaceSchedule(InstSchedule *OldSched, InstSchedule *NewSched,
                              bool IsGlobal);
-  DCF_OPT ParseDCFOpt(std::string opt);
+  DCF_OPT ParseDCFOpt(const std::string &opt);
 
   void PrintPheromone();
 
--- a/include/opt-sched/Scheduler/aco.h
+++ b/include/opt-sched/Scheduler/aco.h
@@ -23,6 +23,13 @@ namespace opt_sched {
 
 typedef double pheromone_t;
 
+enum class DCF_OPT {
+  OFF,
+  GLOBAL_ONLY,
+  GLOBAL_AND_TIGHTEN,
+  GLOBAL_AND_ITERATION
+};
+
 struct Choice {
   SchedInstruction *inst;
   double heuristic;  // range 1 to 2
@@ -45,7 +52,8 @@ private:
   pheromone_t &Pheromone(SchedInstruction *from, SchedInstruction *to);
   pheromone_t &Pheromone(InstCount from, InstCount to);
   double Score(SchedInstruction *from, Choice choice);
-  bool shouldReplaceSchedule(InstSchedule *OldSched, InstSchedule *NewSched);
+  bool shouldReplaceSchedule(InstSchedule *OldSched, InstSchedule *NewSched, bool IsGlobal);
+  DCF_OPT ParseDCFOpt(std::string opt);
 
   void PrintPheromone();
 
@@ -88,6 +96,8 @@ private:
   bool IsPostBB;
   bool IsTwoPassEn;
   pheromone_t ScRelMax;
+  DCF_OPT DCFOption;
+  SPILL_COST_FUNCTION DCFCostFn;
 };
 
 } // namespace opt_sched
--- a/include/opt-sched/Scheduler/aco.h
+++ b/include/opt-sched/Scheduler/aco.h
@@ -86,6 +86,7 @@ private:
   std::unique_ptr<InstSchedule> InitialSchedule;
   bool VrfySched_;
   bool IsPostBB;
+  bool IsTwoPassEn;
   pheromone_t ScRelMax;
 };
 
--- a/include/opt-sched/Scheduler/data_dep.h
+++ b/include/opt-sched/Scheduler/data_dep.h
@@ -629,6 +629,9 @@ private:
   // The schedule's spill cost according to the cost function used
   InstCount spillCost_;
 
+  // The normalized spill cost (absolute Spill Cost - lower bound of spill cost)
+  InstCount NormSpillCost;
+
   // An array of peak reg pressures for all reg types in the schedule
   InstCount *peakRegPressures_;
 
@@ -676,6 +679,8 @@ public:
   InstCount GetExecCost() const;
   void SetSpillCost(InstCount cost);
   InstCount GetSpillCost() const;
+  void SetNormSpillCost(InstCount cost);
+  InstCount GetNormSpillCost() const;
 
   void ResetInstIter();
   InstCount GetFrstInst(InstCount &cycleNum, InstCount &slotNum);
--- a/include/opt-sched/Scheduler/data_dep.h
+++ b/include/opt-sched/Scheduler/data_dep.h
@@ -14,6 +14,7 @@ Last Update:  Mar. 2011
 #include "opt-sched/Scheduler/defines.h"
 #include "opt-sched/Scheduler/sched_basic_data.h"
 #include "llvm/ADT/SmallVector.h"
+#include "llvm/ADT/IndexedMap.h"
 #include <memory>
 
 namespace llvm {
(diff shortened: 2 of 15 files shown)


include/opt-sched/Scheduler/aco.h
```diff
@@ -45,7 +52,9 @@ class ACOScheduler : public ConstrainedScheduler {
   pheromone_t &Pheromone(SchedInstruction *from, SchedInstruction *to);
   pheromone_t &Pheromone(InstCount from, InstCount to);
   double Score(SchedInstruction *from, Choice choice);
-  bool shouldReplaceSchedule(InstSchedule *OldSched, InstSchedule *NewSched);
+  bool shouldReplaceSchedule(InstSchedule *OldSched, InstSchedule *NewSched,
+                             bool IsGlobal);
```
Quincunx271, line 56: `bool` parameters should generally be avoided, as it's hard to tell from the calling code what the bool signifies. I recommend either an `enum class` or using the comment-based named parameters that clang-tidy will check:

```
enum class IS_GLOBAL {
  YES,
  NO,
};

bool shouldReplaceSchedule(..., IS_GLOBAL IsGlobal);
...
shouldReplaceSchedule(Sched, NewSched, IS_GLOBAL::YES);

// Alternatively, in the calling code:
shouldReplaceSchedule(Sched, NewSched, /*IsGlobal=*/true);
// This is checked by clang-tidy, so a typo will be flagged
// (should verify this before relying on it; we might have turned off the check)
```

include/opt-sched/Scheduler/aco.h
```diff
@@ -45,7 +52,9 @@ class ACOScheduler : public ConstrainedScheduler {
...
+  DCF_OPT ParseDCFOpt(std::string opt);
```
Quincunx271, line 57: I'd use `llvm::StringRef` or `const std::string&` instead of `std::string`, which means that the string won't need to be copied into the argument of `ParseDCFOpt()`

include/opt-sched/Scheduler/sched_basic_data.h
```diff
@@ -98,6 +98,9 @@ const int MAX_DEFS_PER_INSTR = 4096;
 // The maximum number of register usages per instruction node.
 const int MAX_USES_PER_INSTR = 4096;
 
+// function for parsing cost function names to enum values
+SPILL_COST_FUNCTION ParseSCFName(std::string name);
```
Quincunx271, line 102: Similarly, `llvm::StringRef` or `const std::string&`

lib/Scheduler/aco.cpp
```diff
@@ -202,13 +242,13 @@ std::unique_ptr<InstSchedule> ACOScheduler::FindOneSchedule() {
   std::unique_ptr<InstSchedule> schedule =
       llvm::make_unique<InstSchedule>(machMdl_, dataDepGraph_, true);
   InstCount maxPriority = rdyLst_->MaxPriority();
-  if (maxPriority == 0)
-    maxPriority = 1; // divide by 0 is bad
+  maxPriority |= !maxPriority; // divide by 0 is bad(increase to 1 if 0)
```
Quincunx271, line 245: I think this is harder to understand than the if statement it replaces.

lib/Scheduler/aco.cpp
```diff
@@ -202,13 +242,13 @@ std::unique_ptr<InstSchedule> ACOScheduler::FindOneSchedule() {
...
   Initialize_();
   rgn_->InitForSchdulng();
 
   SchedInstruction *waitFor = NULL;
   InstCount waitUntil = 0;
+  double maxPriorityInv = 1 / maxPriority;
```
Quincunx271, line 251: Do we know that micro-optimizations like this are helping (i.e. did benchmarks show it)?

Also, I think this can be marked `const`.

lib/Scheduler/bb_spill.cpp
```diff
@@ -313,6 +303,42 @@ InstCount BBWithSpill::CmputCostLwrBound() {
...
+  return RpCostLwrBound_;
+}
+/*****************************************************************************/
+
+void BBWithSpill::addRecordedCost(SPILL_COST_FUNCTION Scf) {
+  // Do nothing if the CF is already recorded
+  for (auto it = recordedCostFunctions.begin();
+       it != recordedCostFunctions.end(); ++it)
+    if (*it == Scf)
+      return;
```
Quincunx271, line 332: [`llvm::is_contained()`](https://llvm.org/doxygen/namespacellvm.html#acd1cd968cb420c82d70926920fcdc7d7).

```
if (llvm::is_contained(recordedCostFunctions, Scf)) return;
```

lib/Scheduler/bb_spill.cpp
```diff
@@ -391,6 +418,11 @@ InstCount BBWithSpill::CmputCost_(InstSchedule *sched, COST_COMP_MODE compMode,
   sched->SetSpillCosts(spillCosts_);
   sched->SetPeakRegPressures(peakRegPressures_);
   sched->SetSpillCost(crntSpillCost_);
+
+  for (auto it = recordedCostFunctions.begin();
+       it != recordedCostFunctions.end(); ++it)
+    storeExtraCost(sched, *it);
```
Quincunx271, line 424:
```
for (const auto& CostFunction : recordedCostFunctions)
  storeExtraCost(sched, CostFunction);
```

include/opt-sched/Scheduler/data_dep.h
```diff
@@ -13,6 +13,7 @@ Last Update:  Mar. 2011
 #include "opt-sched/Scheduler/buffers.h"
 #include "opt-sched/Scheduler/defines.h"
 #include "opt-sched/Scheduler/sched_basic_data.h"
+#include "llvm/ADT/IndexedMap.h"
```
Quincunx271, line 16: Is this `#include` used?

Quincunx271 (changes requested):
There's a few things that I think would improve the quality of the code. The semantic changes look good to me.

include/opt-sched/Scheduler/aco.h
paul-mchugh: This is a private method, that is only used in two places(and cannot conceivably or meaningfully be used elsewhere).  The purpose of this method was to move complicated logic out of a if statement condition.  I don't want to create a enum for so trivial of a purpose.  I will use the `/*IsGlobal=*/` comment.

include/opt-sched/Scheduler/aco.h
paul-mchugh: will do

include/opt-sched/Scheduler/sched_basic_data.h
paul-mchugh: will do

lib/Scheduler/aco.cpp
paul-mchugh: I don't think that this is that hard to understand.  This change removes commonly executed branch from the code, which frees up a slot in the branch predictor tables.  I am open to putting a more descriptive comment if you think that would be appropriate.

lib/Scheduler/aco.cpp
paul-mchugh: This one in particular will.  I profiled ACO a while ago, and the fp division instructions were responsible for a significant portion of the execution time. There are 10 ants in an iteration.  previously there were 2 fp divisions/ant for a total of 20 divisions/iteration.  This change replaces 10 fp divisions with 1 fp division and 10 fp multiplications.

lib/Scheduler/bb_spill.cpp
paul-mchugh: Thanks, I will swap out this code

lib/Scheduler/bb_spill.cpp
paul-mchugh: Thanks for mentioning this.  I will implement this suggestion

include/opt-sched/Scheduler/data_dep.h
paul-mchugh: Nope, I forgot to remove this include.

lib/Scheduler/bb_spill.cpp
paul-mchugh: My one change to your suggested code is that I am not using the const reference to the cost function.  `recordedCostFunctions` stores 32 bit integers (SPILL_COST_FUNCTION is an enum).  Copying a 32 bit integer is no more costly than de-referencing the refrence/pointer to an int(since const refrences to integers are usually converted to int copies by the compiler anyway).
```
  for (SPILL_COST_FUNCTION CostFunction : recordedCostFunctions)
    storeExtraCost(sched, CostFunction);
```

include/opt-sched/Scheduler/sched_basic_data.h
paul-mchugh: I fixed this issue in the latest commit

lib/Scheduler/aco.cpp
paul-mchugh: In light of the information that the compiler optimizes out the branch I have reverted these lines of code.

I know what the asm instructions do.
```asm
test Reg, Reg ; Sets the zero flag in the flags register to 1 if Reg==0, set the zero flag to 1 otherwise
cmovz Reg, 1  ; moves the immediate 1 to Reg if the zero flag == 1
```

```asm
test Reg, Reg ; Sets the zero flag in the flags register to 1 if Reg==0, set the zero flag to 1 otherwise
sete al       ; Sets al to the opposite of of the zero flag (!ZF)
or Reg, al    ; or's Reg with al
```

Quincunx271 (approved):
LGTM

lib/Scheduler/aco.cpp
Quincunx271: Those asm instructions were rough; I didn't actually use a compiler because I was commenting on my phone. For a more accurate view into what the compiler is doing, see https://godbolt.org/z/c9YzvK (I tested it with up to the earliest versions of clang and gcc available there, and it's effectively the same code either way). So it's not exactly the same asm that I wrote (e.g. `cmov`s can't have immediate operands, it seems), but it's still branchless.

PR #143, opened 2021-01-20
Add mode to ddg2dot.py to make differences in edges apparent

Quincunx271:
When edges are added or removed via graph transformations, comparing the
DDGs via ddg2dot.py would be rather difficult. This augments ddg2dot.py
to add invisible edges to encourage `dot` to produce the graphs in the
same layout, with just the visibility of the edges being different.

<details>
<summary>Example graphs augmented with this new feature.</summary>

kernel_c101_sdk_900 18 ddg

kernel_c101_sdk_900 18 occupancy-preserving-ilp nodesup ddg
</details>

### DIFF
--- a/util/misc/ddg2dot.py
+++ b/util/misc/ddg2dot.py
@@ -3,14 +3,10 @@ import argparse
 import sys
 import re
 
-parser = argparse.ArgumentParser(
-    description='Convert data_dep WriteToFile format to a .dot file')
-parser.add_argument(
-    'input', help='The WriteToFile format file to convert. Input a single hyphen (-) to read from stdin')
-parser.add_argument(
-    '-o', '--output', help='The destination to write to. Defaults to stdout')
-parser.add_argument('--filter-weights', nargs='*', default=[],
-                    help='filter out weights with the respective values')
+parser = argparse.ArgumentParser(description='Convert data_dep WriteToFile format to a .dot file')
+parser.add_argument('input', help='The WriteToFile format file to convert. Input a single hyphen (-) to read from stdin')
+parser.add_argument('-o', '--output', help='The destination to write to. Defaults to stdout')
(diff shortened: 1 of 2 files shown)


util/misc/ddg2dot.py
```diff
@@ -30,37 +44,71 @@
...
+            )
+        )
 
 # Graph is now finished:
 result.append('}\n')
 
 output = sys.stdout
-if args.output: output = open(args.output, 'w')
+if args.output:
+    output = open(args.output, 'w')
```
paul-mchugh, line 112: The output file descriptor is not closed by the time the program ends.  This is a nit and the file will be closed anyway when the program exits, so I am okay with not doing anything about this comment.  But resource leaks are usually considered bad form.

paul-mchugh (commented):
I haven't used your ddg2dot.py script, so any changes to this script will not effect my workflows. I have a couple nits and questions.

Why did you split some lines into two lines?  It is my understanding that the 80 character rule only applies to c++.  Where it doesn't apply, I don't think we should split up the lines since I think it makes it harder to read(on larger screens)(I dont think any of us view our code on an 80 char display). eg: see lines 8-15.

My other nit is the is the inline comment on line 112.

Quincunx271:
>Why did you split some lines into two lines?

I now have a Python code formatter (autopep8) installed on my machine, so I applied it to the file. I didn't mess with any of the formatter's settings, and PEP8 prescribes an 80 character line limit, so it wrapped the lines. I tend to agree with you that it's not necessary and somewhat hampers readability, but I'd prefer to leave it formatted according to autopep8 unless we figure some configuration file like that .clang-format file.

Quincunx271:
@paul-mchugh I addressed your comments. I found how to make a configuration file for autopep8 as well, so the unnecessary line-wrapping is reverted now.

paul-mchugh (approved):
LGTM

PR #144, opened 2021-02-10
Update rp graph trans

Quincunx271:
Updates the RP graph transformations. Some bugfixes and refactoring, but also adds some features/refactorings needed by the ILP or RP+ILP graph transformations.

This is part of bringing my branches up to date so that they can be merged into master.


paul-mchugh (approved):
LGTM, I don't have a ton of experience with the graph transformation code, but I looked over the code and don't see any issues.

PR #145, opened 2021-03-04
ILP Graph Transformations

Quincunx271:
Implements (mostly) the ILP graph transformations from Dr. Heffernan's paper.

Not implemented:
 - Subgraph superiority
 - Resource edges (not modeled under our machine model)


VladM1076 (approved):
LGTM.
I am not familiar with the ILP Graph Transformations paper, but I do not see any issues with this commit.

PR #150, opened 2021-03-24
Add reduced latency pass patch

vangthao95:
This patch adds a reduced latency pass that can be used with the two-pass algorithm. This new pass is intended to be done after the Occupancy pass and can be used with or in place of the ILP pass.

The new reduced latency pass can be enabled by modifying the `sched.ini` file and modifying the `PASS_ORDER` setting to include the value `3`. There are also settings to tune the pass such as `LATENCY_TARGETS`, `LATENCY_DIVISOR`, and `LATENCY_MINIMUM`.

Since this pass reduces latencies, we will not get accurate schedule length values so an additional pass can be enabled by setting `COMPILE_TIME_DATA_PASS YES`. This additional pass runs through the schedule again with normal latencies by using the sequential list scheduler. Note that this will increase compile-time due to adding an additional pass. If you do not care about the compile-time data of our scheduler during the reduced latency pass then this pass can be left disabled.

Additionally, in this patch range tightening and relaxed scheduling pruning during enumeration have also been disabled in the Occupancy pass. Since all latencies are 1 in the occupancy pass, we know the min/max length of schedules thus do not need length-related pruning. The lower bound length calculation done before enumerating has also been disabled in any regions with max latency <= 1 since in these cases the lower bound length should be the number of instructions in the region.

### DIFF
--- a/example/llvm7-CPU2006-cfg/sched.ini
+++ b/example/llvm7-CPU2006-cfg/sched.ini
@@ -14,6 +14,32 @@ PRINT_SPILL_COUNTS YES
 # NO
 USE_TWO_PASS NO
 
+# Order of passes to run delimited by a comma with no space. Only enabled with
+# USE_TWO_PASS YES. Note that if both the regular ILP Pass and the ILP Reduced
+# Latency Pass are enabled then only timed out regions in the regular ILP pass
+# will be selected to be re-schedule in the Reduced Latency Pass.
+# VALUES:
+# OCC - Occupancy Pass
+# ILP - Regular ILP Pass
+# ILP_RL - ILP Reduced Latency Pass
+# Valid orderings: OCC,ILP or OCC,ILP_RL or OCC,ILP,ILP_RL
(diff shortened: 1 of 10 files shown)


vangthao95:
Do: Reformat

Quincunx271 (approved):
I went through the code, and everything seems to make sense to me. I would suggest that `PASS_ORDER 1,2` be something like `PASS_ORDER OCC, ILP` (my suggestions: 1 = OCC, 2 = ILP, 3 = ILP_RL), but it's not a big deal either way.

VladM1076 (approved):
LGTM

vangthao95:
> I went through the code, and everything seems to make sense to me. I would suggest that `PASS_ORDER 1,2` be something like `PASS_ORDER OCC, ILP` (my suggestions: 1 = OCC, 2 = ILP, 3 = ILP_RL), but it's not a big deal either way.

I like that idea. I will try to see if I can get that in.

PR #151, opened 2021-03-29
Fix #150 not disabling range tightening

vangthao95:
Commit #150 did not disable range tightening even though it was supposed to.


vangthao95:
Makes sense. I have updated the pull request.

lib/CMakeLists.txt
```diff
@@ -5,3 +5,5 @@ else()
 endif()
 
 add_dependencies(OptSched ${OPT_SCHED_TARGET_DEPS})
+
+target_link_libraries(OptSched -L/home/jeff/ROCm/opencl/build/lib/ libamdocl64.so)
```
Quincunx271, line 9: Needs to be removed for the PR.

However, note that you can do this without modifying the CMakeLists.txt now: when you configure with CMake (or reconfigure), add `-DOPTSCHED_EXTRA_LINK_LIBRARIES=/path/to/ROCm/opencl/build/lib/libamdocl64.so`

lib/Scheduler/enumerator.cpp
```diff
@@ -993,6 +993,10 @@ bool Enumerator::FindNxtFsblBrnch_(EnumTreeNode *&newNode) {
 #endif
 
     if (i == brnchCnt - 1) {
+      if (!(!getIsTwoPass() || getIsSecondPass())) {
+        return false;
+      }
```
Quincunx271, line 998: Recommend De Morgan's-izing this: `if (getIsTwoPass() && !getIsSecondPass()) {`. I think that's a little easier to understand, "If we are in two-pass mode but not in the second pass".

Quincunx271 (changes requested):

lib/Scheduler/enumerator.cpp
Quincunx271: Alternatively, factor out the `(!getIsTwoPass() || getIsSecondPass())` with the other `if` further down, so that this is `if (!Condition)` and the other is `if (Condition)`

lib/CMakeLists.txt
jrbyrnes: Thanks for the tip!

Quincunx271 (approved):
From my tests, there doesn't seem to be any issues with this.

PR #158, opened 2021-06-10
Fix CMake build

Quincunx271:
Label CUDA source files with appropriate extension so that CMake can use
the correct compiler and flags.

Set appropriate variables to do separable compilation.

-----

I did this with the most recent CMake: 3.20.3 . You can find that at /home/justin/cmake-3.20/bin/cmake on the Tesla.


Quincunx271:
Uh, I thought I specifically saw sm_75 in my build.ninja file. I must have forgotten to refresh the build directory when I did that.

> Do you know of a way to force compilation with CC 7.5?

It's actually really easy: configure with `-DCMAKE_CUDA_ARCHITECTURES=75`.

> The CPP files need to include /usr/local/cuda/include

I have no clue why that didn't happen. I assumed that CMake set that up for me, as it compiles perfectly fine for me.

What do you get from `g++ -xc++ -E -v -`? This is what I see in mine:

```
#include "..." search starts here:
#include <...> search starts here:
 /usr/local/cuda-11.3/targets/x86_64-linux/include
 .
 /usr/include/c++/9
 /usr/include/x86_64-linux-gnu/c++/9
 /usr/include/c++/9/backward
 /usr/lib/gcc/x86_64-linux-gnu/9/include
 /usr/local/include
 /usr/include/x86_64-linux-gnu
 /usr/include
End of search list.
```

cuda_runtime.h is in /usr/local/cuda-11.3/targets/x86_64-linux/include. Note that `/usr/local/cuda/include` is, after expanding all the symbolic links, the same directory.

> how would I go about adding the flags "-lineinfo --ptxas-options=-v" to the CUDA files compilation

You'd configure with `-DCMAKE_CUDA_FLAGS_INIT='-lineinfo --ptxas-options=-v'` with a fresh build directory (`CMAKE_CUDA_FLAGS` if the directory isn't fresh). You could also run `ccmake` for a TUI:

 - `ccmake .` in the build directory
 - (t)oggle advanced mode.
 - `/CUDA_FLAGS` if you don't see `CMAKE_CUDA_FLAGS` should move the TUI cursor to the `CMAKE_CUDA_FLAGS` field.
 - `<enter>` to edit the entry, type your flags, `<enter>` to confirm.
 - (c)onfigure
 - (g)enerate

---

My configure: `~/cmake-3.20/bin/cmake -GNinja -DCMAKE_CUDA_ARCHITECTURES=75 -DCMAKE_INSTALL_PREFIX=$FLANG_INSTALL -DCMAKE_BUILD_TYPE=Release -DLLVM_TARGETS_TO_BUILD='X86;AArch64' ..`

VladM1076:
When I do `g++ -xc++ -E -v -` I get:
```
#include "..." search starts here:
#include <...> search starts here:
 /usr/local/cuda-11.3/targets/x86_64-linux/include
 .
 /usr/include/c++/9
 /usr/include/x86_64-linux-gnu/c++/9
 /usr/include/c++/9/backward
 /usr/lib/gcc/x86_64-linux-gnu/9/include
 /usr/local/include
 /usr/include/x86_64-linux-gnu
 /usr/include
 ```

Strange that the cpp files don't link the cuda includes properly since we both have `/usr/local/cuda-11.3/targets/x86_64-linux/include`

I will try this again later today once the test I am running on the Tesla completes. Thank you for the very informative answer! This will make building the cuda code much simpler.

VladM1076 (approved):
It works!
For the record the command used was `/home/justin/cmake-3.20/bin/cmake -GNinja -DCMAKE_CUDA_ARCHITECTURES=75 -DCMAKE_CUDA_FLAGS_INIT='-lineinfo --ptxas-options=-v' -DCMAKE_CXX_FLAGS_INIT='-isystem/usr/local/cuda/include' -DCMAKE_INSTALL_PREFIX=$FLANG_INSTALL -DCMAKE_BUILD_TYPE=Release -DLLVM_TARGETS_TO_BUILD='X86;AArch64' ..`
Once cmake 3.20 is installed system wide, specific cmake will not need to be specified.
Good work Justin!

PR #162, opened 2021-06-25
Small Refactor of MachineModel to facilitate tests

Quincunx271:
By allowing MachineModel to be constructed by a SpecsBuffer rather than a file, we can easily get a MachineModel in any tests that have one as a dependency.

---

There are bigger ways to refactor the MachineModel that would definitely improve things, but if the focus is making the minimal change needed to enable more widespread unit tests, this is probably the simplest possible change.


Quincunx271:
Helps enable #177

PR #167, opened 2021-07-07
GT analysis scripts

Quincunx271:
Adds graph transformation analysis scripts. As part of this, several library analysis functions are added.

This is probably not ready to merge yet


Quincunx271:
This is a massive PR with a lot of changes. It would be reasonable to split this into many smaller PRs if that would make it easier to review. Please tell me if I should go through with that; I don't want to do the work to split it if it's not going to be useful.

PR #178, opened 2022-04-12
Fix master bug

jrbyrnes:
Hey --

Currently the master branch has a runtime issue from one of the last few commits. I don't really have the time to fully investigate why the bug is occurring, but the error is essentially that SetupForSchduling uses `issueTypes_.size()` before it is initialized. I believe this has something to do with a change in the way the machine model is constructed, but I am not sure.

This PR forces initialization of the vector by loading in the MM config file. If anyone knows off the top of their head what the culprit is, I can modify, otherwise I'll merge soon.

The key idea here is that any new team members won't need to debug the code before they are able to set up their environments.

### DIFF
--- a/lib/Scheduler/machine_model.cpp
+++ b/lib/Scheduler/machine_model.cpp
@@ -14,37 +14,7 @@ using namespace llvm::opt_sched;
 using std::string;
 using std::vector;
 
-MachineModel::MachineModel(const std::string &modelFile) {
-  SpecsBuffer buf;
-  buf.Load(modelFile.c_str());
-
-  char buffer[MAX_NAMESIZE];
-
-  buf.ReadSpec("MODEL_NAME:", buffer);
-  mdlName_ = buffer;
-
-  issueRate_ = buf.ReadIntSpec("ISSUE_RATE:");
-
-  int numIssueTypes = buf.ReadIntSpec("ISSUE_TYPE_COUNT:");
-
-  issueTypes_.resize(numIssueTypes > 0 ? numIssueTypes : 1);
-  if (numIssueTypes > 0) {
-    for (size_t j = 0; j < issueTypes_.size(); j++) {
-      int pieceCnt;
-      char *strngs[INBUF_MAX_PIECES_PERLINE];
-      int lngths[INBUF_MAX_PIECES_PERLINE];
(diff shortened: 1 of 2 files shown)


jrbyrnes:
Do: Reformat

lib/Scheduler/machine_model.cpp
```diff
@@ -17,6 +17,31 @@ using std::vector;
...
+      int lngths[INBUF_MAX_PIECES_PERLINE];
+      buf.GetNxtVldLine(pieceCnt, strngs, lngths);
+
+      if (pieceCnt != 2)
+        llvm::report_fatal_error("Invalid issue type spec", false);
+
+      issueTypes_[j].name = strngs[0];
+      issueTypes_[j].slotsCount = atoi(strngs[1]);
+    }
+  }
```
Quincunx271, line 44: Based on the change that introduced this bug (#162), I think this should include all the other code of the other constructor as well.

Rather than duplicate the code below, I would suggest pulling the other constructor's body into a helper method and having this constructor call into it:

```c++
// Member of MachineModel:
private:
  void parseBuffer(SpecsBuffer &buf);

// In machine_model.cpp:
void MachineModel::parseBuffer(SpecsBuffer &buf) {
  // The other constructor body.
}

// The two constructors would forward to that:
MachineModel::MachineModel(const std::string& modelFile) {
  SpecsBuffer buf;
  buf.Load(modelFile.c_str());
  parseBuffer(buf);
}

MachineModel::MachineModel(SpecsBuffer &buf) {
  parseBuffer(buf);
}
```

Quincunx271 (changes requested):
Nice catch. #162 apparently tried to split off another constructor, but it looks like it forgot to forward this constructor on to the other.

lib/Scheduler/machine_model.cpp
jrbyrnes: I'm not sure why I didn't see these comments when you posted -- sorry about that. Anyway, I agree with this design. I'll update the commit

jrbyrnes:
Do: Reformat

Quincunx271 (approved):
LGTM. Please rerun the checks with a comment of `Do: Run Checks` .

jrbyrnes:
Do: Run Checks

PR #181, opened 2022-06-28
Support using a custom value for occupancy limit across all kernels

jrbyrnes:
This PR implements some logic to bring in support for using `OCCUPANCY_LIMIT` to set a custom limit of occupancy for all kernels. To use this option, set `OCCUPANCY_LIMIT_SOURCE` to `VALUE`


JoshHuttonCode (approved):
LGTM

